PR #62, opened 2024-09-24
feat: Enable Sidecar to detect SR running on default local port

ringuva:
## Summary of Changes

Enable Sidecar to detect Schema Registry (SR) running on default port and use it to decode messages when consuming from local Kafka topic. Issue [156]( https://github.com/confluentinc/vscode/issues/156)

- SR default port on local cluster is 8081. [link](https://docs.confluent.io/platform/current/schema-registry/schema_registry_onprem_tutorial.html)
- Configuration of SR port will be handled in a subsequent PR.
- Fixed flaky system test. The flakiness was due to repeated start & stop of kafka & SR containers for each test.

## Pull request checklist

- Tests:
    - [x] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java
@@ -28,11 +28,11 @@ public record ConnectionSpec(
   }
 
   public ConnectionSpec withId(String id) {
-    return new ConnectionSpec(id, name, type, ccloudConfig, null);
+    return new ConnectionSpec(id, name, type, ccloudConfig, localConfig);
   }
 
   public ConnectionSpec withName(String name) {
-    return new ConnectionSpec(id, name, type, ccloudConfig, null);
+    return new ConnectionSpec(id, name, type, ccloudConfig, localConfig);
   }
 
   /**
@@ -40,7 +40,7 @@ public record ConnectionSpec(
    * Confluent Cloud organization ID set in the CCloudConfig.
    */
   public ConnectionSpec withCCloudOrganizationId(String ccloudOrganizationId) {
-    return new ConnectionSpec(id, name, type, new CCloudConfig(ccloudOrganizationId), null);
+    return new ConnectionSpec(id, name, type, new CCloudConfig(ccloudOrganizationId), localConfig);
   }
 
   public String ccloudOrganizationId() {
@@ -58,9 +58,6 @@ public record ConnectionSpec(
   public record LocalConfig(
       @JsonProperty(value = "schema-registry-uri") String schemaRegistryUri
   ) {
-    public String schemaRegistryUri() {
-      return this.schemaRegistryUri;
-    }
   }
 
   /**
--- a/src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java
@@ -13,7 +13,8 @@ public record ConnectionSpec(
     String id,
     String name,
     ConnectionType type,
-    @JsonProperty("ccloud_config") CCloudConfig ccloudConfig
+    @JsonProperty("ccloud_config") CCloudConfig ccloudConfig,
+    @JsonProperty("local_config") LocalConfig localConfig
 ) {
 
   public enum ConnectionType {
@@ -23,15 +24,15 @@ public record ConnectionSpec(
   }
 
   public ConnectionSpec(String id, String name, ConnectionType type) {
-    this(id, name, type, null);
+    this(id, name, type, null, null);
   }
 
   public ConnectionSpec withId(String id) {
-    return new ConnectionSpec(id, name, type, ccloudConfig);
+    return new ConnectionSpec(id, name, type, ccloudConfig, null);
   }
 
   public ConnectionSpec withName(String name) {
-    return new ConnectionSpec(id, name, type, ccloudConfig);
+    return new ConnectionSpec(id, name, type, ccloudConfig, null);
   }
 
   /**
@@ -39,7 +40,7 @@ public record ConnectionSpec(
    * Confluent Cloud organization ID set in the CCloudConfig.
    */
   public ConnectionSpec withCCloudOrganizationId(String ccloudOrganizationId) {
-    return new ConnectionSpec(id, name, type, new CCloudConfig(ccloudOrganizationId));
+    return new ConnectionSpec(id, name, type, new CCloudConfig(ccloudOrganizationId), null);
   }
 
   public String ccloudOrganizationId() {
@@ -51,6 +52,15 @@ public record ConnectionSpec(
       @JsonProperty(value = "organization_id", required = true) String organizationId
   ) {
 
+  }
+
+  @Schema(description = "Configuration for local cluster")
+  public record LocalConfig(
+      @JsonProperty(value = "schema-registry-uri") String schemaRegistryUri
+  ) {
+    public String schemaRegistryUri() {
+      return this.schemaRegistryUri;
+    }
   }
 
   /**
--- a/src/main/java/io/confluent/idesidecar/restapi/models/graph/FakeLocalFetcher.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/models/graph/FakeLocalFetcher.java
@@ -34,6 +34,6 @@ public class FakeLocalFetcher implements LocalFetcher {
 
   @Override
   public Uni<LocalSchemaRegistry> getSchemaRegistry(String connectionId) {
-    return Uni.createFrom().item(new LocalSchemaRegistry("http://localhost:8081", connectionId));
+    return Uni.createFrom().item(new LocalSchemaRegistry(connectionId, "http://localhost:8081"));
   }
 }
--- a/src/main/java/io/confluent/idesidecar/restapi/models/graph/FakeLocalFetcher.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/models/graph/FakeLocalFetcher.java
@@ -34,6 +34,6 @@ public class FakeLocalFetcher implements LocalFetcher {
 
   @Override
   public Uni<LocalSchemaRegistry> getSchemaRegistry(String connectionId) {
-    return Uni.createFrom().item(new LocalSchemaRegistry(connectionId));
+    return Uni.createFrom().item(new LocalSchemaRegistry("http://localhost:8081", connectionId));
   }
 }
--- a/src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
@@ -152,17 +152,23 @@ public class RealLocalFetcher extends ConfluentLocalRestClient implements LocalF
     }
   }
 
-  public Uni<LocalSchemaRegistry> getSchemaRegistry(
-      String connectionId
-  ) {
+  public String resolveSchemaRegistryUri(String connectionId) {
     var localConfig = connections.getConnectionSpec(connectionId).localConfig();
+    // Null -> discover local SR, Blank -> no SR (return null for blank)
     String uri = Optional.ofNullable(localConfig)
         .map(LocalConfig::schemaRegistryUri)
-        .filter(url -> !url.isBlank())
         .orElse(CONFLUENT_LOCAL_DEFAULT_SCHEMA_REGISTRY_URI);
+    if (uri.isBlank()) {
+      Log.warnf("Schema Registry URI is blank for connection: %s", connectionId);
+      return null;
+    }
+    return uri;
+  }
 
-    if (uri == null || uri.isBlank()) {
-      Log.warnf("Schema Registry URL is missing or blank for connection: %s", connectionId);
+
+  public Uni<LocalSchemaRegistry> getSchemaRegistry(String connectionId) {
+    String uri = resolveSchemaRegistryUri(connectionId);
+    if (uri == null) {
       return Uni.createFrom().nullItem();
     }
 
--- a/src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
@@ -153,15 +153,22 @@ public class RealLocalFetcher extends ConfluentLocalRestClient implements LocalF
   public String resolveSchemaRegistryUri(String connectionId) {
     var localConfig = connections.getConnectionSpec(connectionId).localConfig();
     if (localConfig == null || localConfig.schemaRegistryUri() == null) {
+      // Use the default SR URI when missing LocalConfig or missing SR URI in that config
       return CONFLUENT_LOCAL_DEFAULT_SCHEMA_REGISTRY_URI;
     }
-    return localConfig.schemaRegistryUri(); // may be null if no SR is to be used
+    var uri = localConfig.schemaRegistryUri();
+    if (uri.isBlank()) {
+      // Blank URL means no SR
+      return null;
+    }
+    return uri;
   }
 
 
   public Uni<LocalSchemaRegistry> getSchemaRegistry(String connectionId) {
     String uri = resolveSchemaRegistryUri(connectionId);
     if (uri == null) {
+      Log.infof("No Schema Registry will be used for connection %s", connectionId);
       return Uni.createFrom().nullItem();
     }
 
--- a/src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
@@ -176,7 +176,7 @@ public class RealLocalFetcher extends ConfluentLocalRestClient implements LocalF
             Log.infof("Request to schema-registry '%s' failed.", configUri);
             return Uni.createFrom().nullItem();
           }
-          return Uni.createFrom().item(() -> new LocalSchemaRegistry(uri, connectionId));
+          return Uni.createFrom().item(() -> new LocalSchemaRegistry(connectionId, uri));
         })
         .map(localSchemaRegistry -> onLoad(connectionId, localSchemaRegistry));
   }
--- a/src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
@@ -11,6 +11,8 @@ import io.confluent.idesidecar.restapi.exceptions.ConnectionNotFoundException;
 import io.confluent.idesidecar.restapi.exceptions.ResourceFetchingException;
 import io.confluent.idesidecar.restapi.models.Connection;
 import io.confluent.idesidecar.restapi.models.ConnectionSpec.ConnectionType;
+import io.confluent.idesidecar.restapi.models.ConnectionSpec.LocalConfig;
+import io.quarkus.logging.Log;
 import io.quarkus.runtime.annotations.RegisterForReflection;
 import io.smallrye.mutiny.Uni;
 import jakarta.enterprise.context.ApplicationScoped;
@@ -19,6 +21,7 @@ import jakarta.inject.Inject;
 import java.net.ConnectException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.regex.Pattern;
 import org.eclipse.microprofile.config.ConfigProvider;
 
@@ -72,6 +75,13 @@ public class RealLocalFetcher extends ConfluentLocalRestClient implements LocalF
       .getConfig()
       .getValue(
           "ide-sidecar.connections.confluent-local.resources.broker-adv-listeners-config-uri",
+          String.class
+      );
+
+  static final String CONFLUENT_LOCAL_DEFAULT_SCHEMA_REGISTRY_URI = ConfigProvider
+      .getConfig()
+      .getValue(
+          "ide-sidecar.connections.confluent-local.default.schema-registry-uri",
           String.class
       );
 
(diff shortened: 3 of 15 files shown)


src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java
```diff
@@ -53,6 +54,15 @@ public record CCloudConfig(
...
   }
 
+  @Schema(description = "Configuration for local cluster")
+  public record LocalConfig(
+      @JsonProperty(value = "schema-registry-uri") String schemaRegistryUri
+  ) {
+    public String schemaRegistryUri() {
+      return this.schemaRegistryUri;
+    }
+  }
```
rhauch, line 64: Why do we want to use a URL here, rather than a port or multiple ports? If it's a _local connection_, then should we assume the schema registry is also running locally?

src/main/java/io/confluent/idesidecar/restapi/models/graph/FakeLocalFetcher.java
```diff
@@ -34,6 +34,6 @@ public Uni<ConfluentLocalKafkaCluster> getKafkaCluster(String connectionId) {
 
   @Override
   public Uni<LocalSchemaRegistry> getSchemaRegistry(String connectionId) {
-    return Uni.createFrom().item(new LocalSchemaRegistry(connectionId));
+    return Uni.createFrom().item(new LocalSchemaRegistry("http://localhost:8081", connectionId));
```
rhauch, line 37: Most of the other types are instantiated in code that doesn't know the connection (most are parsed from JSON responses with external services), and then we use the `withConnectionId(id)` to add that later. So there's no much precedence for passing the connection ID in right away. 

Having said that, would this be a little more clear if the connection ID were first, since most of the methods in the *Fetcher classes take the connection as the first parameter?

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
```diff
@@ -142,11 +152,37 @@ Uni<ConfluentLocalKafkaCluster> withBrokerDetails(ConfluentLocalKafkaCluster clu
...
+    var localConfig = connections.getConnectionSpec(connectionId).localConfig();
+    String uri = Optional.ofNullable(localConfig)
+        .map(LocalConfig::schemaRegistryUri)
+        .filter(url -> !url.isBlank())
+        .orElse(CONFLUENT_LOCAL_DEFAULT_SCHEMA_REGISTRY_URI);
+
+    if (uri == null || uri.isBlank()) {
+      Log.warnf("Schema Registry URL is missing or blank for connection: %s", connectionId);
+      return Uni.createFrom().nullItem();
+    }
```
rhauch, line 167: So with this logic the `uri` on line 159 will never be null or blank, and lines 164-167 are not really needed. 

But is that what we want? Seems like this method should return `Uni.createfrom().nullItem()` when the connection specifies a null or blank SR URI.

IOW, when do (or should) we use the default specified in `CONFLUENT_LOCAL_DEFAULT_SCHEMA_REGISTRY_URI`? I guess some of this depends on how long it takes for our request to discover whether there is an SR at the specified or default URI -- if it's too long, then we might not want to default to anything.

One option is to do exactly what the Connection specifies: if it's there, use it; if not and it's optional, don't. Another is to always look for a schema registry that is either what's specified or the default; IMO this must be fast and must not introduce any delay, since there is no way for me to say "I'm not using SR".

Thoughts?

src/test/java/io/confluent/idesidecar/restapi/resources/ConnectionsResourceTest.java
```diff
@@ -464,7 +464,7 @@ void updateConnectionWithMultipleValidationErrors() {
 
     var badSpec = new ConnectionSpec(
         "c3", "Connection name changed!", ConnectionType.PLATFORM,
-        new CCloudConfig("org-id"));
+        new CCloudConfig("org-id"), null);
```
rhauch, line 467: Nit: proper indentation and line feeds helps _a lot_ here:
```suggestion
        new CCloudConfig("org-id"),
        null
    );
```

src/test/java/io/confluent/idesidecar/restapi/util/CCloudTestUtil.java
```diff
@@ -228,7 +228,8 @@ public ConnectionSpec createConnection(
                 connectionId,
                 connectionName,
                 connectionType,
-                ccloudConfig))
+                ccloudConfig,
+                null))
```
rhauch, line 232: Nit:
```suggestion
                null
            )
        )
```

rhauch (commented):
Thanks, @ringuva. A few comments/questions below.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
rhauch: Okay, after thinking about it a bit, I think we need to first decide from the Connection API perspective what it means for `local_config.schema_registry_uri` to be null and when it is blank.
1. Null means "discover a local SR" and blank means "no SR"; or
2. Null and blank both mean "no SR".
3. Null or blank both mean "discover a local SR"

where "discover a local SR" could mean multiple things, including the initial behavior of "look at http://localhost:8081".

This PR is essentially implementing option 3. IMO, this seems weird in the long run. It *might* make sense in the short term, before the extension changes to specify `local_config` in the `Connection` resource.

But honestly, I think we should maybe choose option 1, and have the extension set the `Connection` resource accordingly -- and the extension UX might need to make it very clear how the user says "I don't have an SR".

Once we decide the API behavior, then we can decide whether it's better to have the default logic in `RealLocalFetcher`, where we have more opportunity to change the "discovery" logic.

ringuva: Why would an user choose 'no SR'  when it is available ?  If the SR is not available or the user couldn't start a SR then searching for a local SR shouldn't be an issue.  I think, we should encourage the users to use SR is if it is available.

Won't having too many options would make it difficult to remember for the API user.  Should the user care about Null Vs Empty ?  If we want to allow the use to ignore SR, then we should expose it as separate option in connection API.

In short, I prefer option-3.  cc @rhauch ?

src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java
ringuva: We are using the URI for the kafka-rest.  See application.yml  `kafkarest-uri`.  This is consistent with a similar option in the section of applciation.yml file.

Should we allow the user to point to an SR that is not running locally but on a different host? Why not? I don't see any strong benefits or significant downsides to disallowing it. @rhauch

src/main/java/io/confluent/idesidecar/restapi/models/graph/FakeLocalFetcher.java
ringuva: I agree, connection as the first parameter is better. 
Done.

src/test/java/io/confluent/idesidecar/restapi/resources/ConnectionsResourceTest.java
ringuva: done.

src/test/java/io/confluent/idesidecar/restapi/util/CCloudTestUtil.java
ringuva: Done.

src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java
rhauch: Sounds good -- I just wanted to be intentional here.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
rhauch: Option 3 is probably the least attractive to me, since it does not allow users to fully control the behavior.

I prefer option 1 since it's simple and easy to distinguish between null and a blank string. You raise a good point about the overall UX, but I think there's a difference between defining the best overall user experience (UX) for the extension+sidecar (which should by default try to discover the local SR), versus designing a good API for the sidecar that makes sense and gives control.

I agree with you that the overall UX probably should try to *discover a local SR*, since that simplifies things for the user. We can do this with option 1 and option 3, but probably not with option 2. The difference between option 1 and option 3 is simply whether the client can ever specify "don't even bother looking for a local SR" -- option 1 allows this, but option 3 does not. I think it's better to allow it with option 1.

src/test/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcherTest.java
```diff
@@ -191,4 +195,42 @@ void assertExtractLocalhostAddresses(String input, String...expectedAddresses) {
...
+
+  @Test
+  void shouldReturnNullWhenSchemaRegistryUriIsBlank() throws CreateConnectionException {
+    var localConfig = new ConnectionSpec.LocalConfig("");
+    var connectionSpec = new ConnectionSpec("3", "Local Connection", ConnectionSpec.ConnectionType.LOCAL, null, localConfig);
+    manager.createConnectionState(connectionSpec);
+
+    String uri = localFetcher.resolveSchemaRegistryUri("3");
+    assertNull(uri); // No Schema Registry
+  }
```
rhauch, line 235: We should also have a test for when the SR URI is null
```suggestion
  }

  @Test
  void shouldReturnDefaultSchemaRegistryUriWhenSchemaRegistryUriIsNull() throws CreateConnectionException {
    var localConfig = new ConnectionSpec.LocalConfig(null);
    var connectionSpec = new ConnectionSpec("4", "Local Connection", ConnectionSpec.ConnectionType.LOCAL, null, localConfig);
    manager.createConnectionState(connectionSpec);

    String uri = localFetcher.resolveSchemaRegistryUri("4");
    assertEquals("http://localhost:8081", uri);
  }
```

src/test/java/io/confluent/idesidecar/restapi/resources/KafkaConsumeResourceIT.java
```diff
@@ -83,15 +105,14 @@ void testConfluentLocalContainer() {
           .body().asString();
 
       assertConsumerRecords(rows, sampleRecords);
-    }
+      deleteLocalConnection(connectionId);
```
rhauch, line 108: This connection will not be removed if the tests fail. There are two options:
1. Add an `@AfterEach` method that removes all connections; or
2. Add an `@AfterEach` method that removes all connections that were created in `setupTestEnvironment(...) -- probably would require bookkeeping the created connection IDs in a field`.

The first would be the easiest with least bookkeeping, but if that's not feasible for some reason then the second should be doable.

src/test/java/io/confluent/idesidecar/restapi/resources/KafkaConsumeResourceIT.java
```diff
@@ -424,7 +438,101 @@ void testConsumeWithMessageSizeLimitAcrossPartitions() {
...
+        MyMessage.Builder messageBuilder = MyMessage.newBuilder();
+        JsonFormat.parser().merge(jsonValue, messageBuilder); // Parse JSON to MyMessage
+        MyMessage parsedMessage = messageBuilder.build();
+
+        // Compare the parsed message with the original message
+        assertEquals(messages.get(i), parsedMessage, "Mismatched Protobuf message");
+
+        assertFalse(newRecords.get(i).get("exceeded_fields").get("value").asBoolean(), "Exceeded fields should be false");
+      }
+    deleteLocalConnection(connectionId);
```
rhauch, line 526: Indentation here is weird.

src/test/java/io/confluent/idesidecar/restapi/resources/KafkaConsumeResourceIT.java
```diff
@@ -496,9 +608,13 @@ void createLocalKafkaTopic(String connectionId,
                 partitionsCount
             )
         )
-        .post(String.format("/kafka/v3/clusters/%s/topics", localKafkaClusterId))
-        .then()
-        .statusCode(201);
+        .post(String.format("/kafka/v3/clusters/%s/topics", localKafkaClusterId));
+    if (response.statusCode() != 201) {
+      System.err.println("Failed to create Kafka topic. Status code: " + response.statusCode());
+      System.err.println("Response body: " + response.getBody().asString());
```
rhauch, line 614: Do we need these lines? If so, can we use something other than `System.err` here?

rhauch (commented):
Thanks again, @ringuva. I have a few more questions and suggestions.

src/test/java/io/confluent/idesidecar/restapi/resources/KafkaConsumeResourceIT.java
ringuva: Done.

src/test/java/io/confluent/idesidecar/restapi/resources/KafkaConsumeResourceIT.java
ringuva: Done.

src/test/java/io/confluent/idesidecar/restapi/resources/KafkaConsumeResourceIT.java
ringuva: Done.

src/test/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcherTest.java
ringuva: Done.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
```diff
@@ -142,11 +152,43 @@ Uni<ConfluentLocalKafkaCluster> withBrokerDetails(ConfluentLocalKafkaCluster clu
...
+
+    final String configUri = uri + "/config";
+    return getItem(connectionId, configUri, this::parseSchemaRegistryConfig)
+        .onFailure(ConnectException.class).recoverWithItem(throwable -> null)
+        .onItem()
+        .transformToUni(response -> {
+          // Verify the compatibility level exists
+          if (response == null || response.compatibilityLevel() == null) {
+            Log.infof("Request to schema-registry '%s' failed.", configUri);
```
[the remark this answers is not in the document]
ringuva: Done.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
```diff
@@ -142,11 +152,43 @@ Uni<ConfluentLocalKafkaCluster> withBrokerDetails(ConfluentLocalKafkaCluster clu
...
+
+
   public Uni<LocalSchemaRegistry> getSchemaRegistry(String connectionId) {
-    // NOTE: Confluent CLI does not support "local" SR yet. It does have a
-    //       `confluent local services schema-registry` command, but this requires
-    //       Confluent Platform to be installed and configured locally.
-    return Uni.createFrom().nullItem();
+    String uri = resolveSchemaRegistryUri(connectionId);
+    if (uri == null) {
+      return Uni.createFrom().nullItem();
```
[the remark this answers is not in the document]
ringuva: Done.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
```diff
@@ -142,11 +152,43 @@ Uni<ConfluentLocalKafkaCluster> withBrokerDetails(ConfluentLocalKafkaCluster clu
...
+    }
+
```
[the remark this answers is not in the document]
ringuva: Done.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
```diff
@@ -142,11 +152,43 @@ Uni<ConfluentLocalKafkaCluster> withBrokerDetails(ConfluentLocalKafkaCluster clu
...
   }
 
+  public String resolveSchemaRegistryUri(String connectionId) {
+    var localConfig = connections.getConnectionSpec(connectionId).localConfig();
+    // Null -> discover local SR, Blank -> no SR (return null for blank)
+    String uri = Optional.ofNullable(localConfig)
+        .map(LocalConfig::schemaRegistryUri)
+        .orElse(CONFLUENT_LOCAL_DEFAULT_SCHEMA_REGISTRY_URI);
+    if (uri.isBlank()) {
+      Log.warnf("Schema Registry URI is blank for connection: %s", connectionId);
```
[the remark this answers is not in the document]
ringuva: Done.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
```diff
@@ -142,11 +152,43 @@ Uni<ConfluentLocalKafkaCluster> withBrokerDetails(ConfluentLocalKafkaCluster clu
     }
   }
 
+  public String resolveSchemaRegistryUri(String connectionId) {
+    var localConfig = connections.getConnectionSpec(connectionId).localConfig();
+    // Null -> discover local SR, Blank -> no SR (return null for blank)
+    String uri = Optional.ofNullable(localConfig)
+        .map(LocalConfig::schemaRegistryUri)
+        .orElse(CONFLUENT_LOCAL_DEFAULT_SCHEMA_REGISTRY_URI);
+    if (uri.isBlank()) {
```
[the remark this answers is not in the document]
ringuva: Done.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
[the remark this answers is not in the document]
rhauch: I don't see this suggestion.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
```diff
@@ -142,11 +150,39 @@ Uni<ConfluentLocalKafkaCluster> withBrokerDetails(ConfluentLocalKafkaCluster clu
     }
   }
 
+  public String resolveSchemaRegistryUri(String connectionId) {
+    var localConfig = connections.getConnectionSpec(connectionId).localConfig();
+    if (localConfig == null || localConfig.schemaRegistryUri() == null) {
+      return CONFLUENT_LOCAL_DEFAULT_SCHEMA_REGISTRY_URI;
+    }
+    return localConfig.schemaRegistryUri(); // may be null if no SR is to be used
```
rhauch, line 158: Okay, my suggestion didn't properly handle a blank URI by returning null. I think we need that, right?

```suggestion
    if (localConfig == null || localConfig.schemaRegistryUri() == null) {
      // Use the default SR URI when missing LocalConfig or missing SR URI in that config
      return CONFLUENT_LOCAL_DEFAULT_SCHEMA_REGISTRY_URI;
    }
    var uri = localConfig.schemaRegistryUri();
    if (uri.isBlank()) {
      // Blank URL means no SR
      return null;
    }
    return uri;
```

src/test/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcherTest.java
```diff
@@ -191,4 +195,51 @@ void assertExtractLocalhostAddresses(String input, String...expectedAddresses) {
...
+
+  @Test
+  void shouldReturnNullWhenSchemaRegistryUriIsBlank() throws CreateConnectionException {
+    var localConfig = new ConnectionSpec.LocalConfig("");
+    var connectionSpec = new ConnectionSpec("3", "Local Connection", ConnectionSpec.ConnectionType.LOCAL, null, localConfig);
+    manager.createConnectionState(connectionSpec);
+
+    String uri = localFetcher.resolveSchemaRegistryUri("3");
+    assertEquals("", uri); // No Schema Registry
```
rhauch, line 234: I understood this `resolveSchemaRegistryUri()` method to get the URI schema registry that we would use, or null if there is none. After all, that's what [this method is doing](https://github.com/confluentinc/ide-sidecar/pull/62/files#diff-7822e65aad428c179cdf0d984e6523c78578616f77d2e6c162cc937caf258eabR164).

I think this test is wrong because my suggestion for the `resolveSchemaRegistryUri()` method was wrong (see earlier comment in this review).

```suggestion
    String uri = localFetcher.resolveSchemaRegistryUri("3");
    assertNul(uri); // No Schema Registry
```

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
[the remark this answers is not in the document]
ringuva: Resolving conflicts at airlock might have removed it. :(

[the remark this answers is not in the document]
ringuva: Resolving conflicts at airlock repo might have removed it.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
ringuva: Done.

src/test/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcherTest.java
ringuva: Done.

src/test/java/io/confluent/idesidecar/restapi/resources/KafkaConsumeResourceIT.java
```diff
@@ -534,6 +534,15 @@ void deleteLocalConnection(String id) {
...
 
+  void deleteLocalConnection(String id) {
+    given()
+        .when()
+        .header("Content-Type", "application/json")
+        .delete("/gateway/v1/connections/" + id)
+        .then()
+        .statusCode(204);
+  }
+
```
rhauch, line 545: This is not needed, right?

ringuva: Only One copy of it is needed ...

rhauch (approved):
LGTM. Thanks for working through this, @ringuva

PR #64, opened 2024-09-25
feat: Make settings related to SSL certificates configurable via Pref…

### LINKED ISSUE #63: Make settings related to SSL certificates configurable via API endpoint
A few users of [Confluent for VS Code](https://github.com/confluentinc/vscode) (https://github.com/confluentinc/vscode/issues/280 https://github.com/confluentinc/vscode/issues/268) reported that they ran into errors when working with Confluent Cloud and sitting behind a firewall that performs SSL inspection.

The sidecar should provide an API endpoint that allows the extension to pass the user's preferences on SSL certificates to the sidecar. Specifically, it should offer options to configure a custom CA cert and to disable verification of certs. The sidecar should respect the preferences and pass them to [instances of the Vert.x web client](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/util/WebClientFactory.java). The sidecar should also provide an API endpoint to read the current preferences.

The endpoint for updating the preferences on the SSL certificates might look as follows:
```
PUT /gateway/v1/preferences

{
  "api_version": "gateway/v1",
  "kind": "Preferences",
  "spec": {
    "tls_pem_paths": ["/Users/stefan/rootca.pem"],
    "trust_all_certificates": false
  }
}
```

The field `tls_pem_paths` can be used to point the sidecar to one or multiple PEM files on the local file system of the user's machine. The field `trust_all_certificates` can be used to enable/disable verification of SSL certificates (if set to `true`, verification is enabled; if set to `false`, verification is disabled). If any of the fields is empty, the sidecar will fallback to the default behavior, which enforces verification of certificates in case of `trust_all_certificates` (i.e., it's set to `false`) and doesn't use a custom certificate in case of `tls_pem_paths`.

The endpoint for reading the current preferences might look as follows:
```
GET /gateway/v1/preferences

Content-Type: application/json
{
  "api_version": "gateway/v1",
  "kind": "Preferences",
  "metadata": {
    "self": "http://localhost:1234/gateway/v1/preferences"
  },
  "spec": {
    "tls_pem_paths": ["/Users/stefan/rootca.pem"],
    "trust_all_certificates": false
  },
  "status": {
    "example_status_field": true
  }
}
```

The field `status` is optional and will not be included if it is empty.

In the future, Confluent for VS Code might use these API endpoints to manage other user preferences, too.

flippingbits:
## Summary of Changes

A few users of [Confluent for VS Code](https://github.com/confluentinc/vscode) (https://github.com/confluentinc/vscode/issues/280 https://github.com/confluentinc/vscode/issues/268) reported that they ran into errors when working with Confluent Cloud and sitting behind a firewall that performs SSL inspection.

This change introduces the Preferences API that can be used to store user preferences in the sidecar.

In the beginning, it supports two preferences related to SSL handshaking:
* The preference `tls_pem_paths` can be used to point the sidecar to one or multiple custom certificates in the PEM format (by default, no custom certificate is configured).
* The preference `trust_all_certificates` can be used to enable/disable the verification of the server's certificate (by default, verification is enabled).

Whenever the preferences change, a [CDI event](https://quarkus.io/guides/cdi#events-and-observers) gets produced so that beans relying on the preferences get notified. This change updates the `WebClientFactory` to `@Observe` the events related to the Preferences API so that the configuration of the web client is updated whenever user change their preferences. In the future, other beans consuming the preferences may `@Observe` the events, too.

Fixes #63
## Pull request checklist

- Tests:
    - [x] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/main/java/io/confluent/idesidecar/restapi/models/Preferences.java
```diff
@@ -0,0 +1,90 @@
...
+      var errors = validateTlsPemPaths();
+
+      if (!errors.isEmpty()) {
+        throw new InvalidPreferencesException(errors);
+      }
+    }
+
+    List<Error> validateTlsPemPaths() {
+      return this.tlsPemPaths.stream()
+          .filter(pemPath -> Files.notExists(Path.of(pemPath)))
```
rhauch, line 71: One thing to consider in the future is whether to also validate that the existing file has well-known extensions, like `.pem`, `.crt`, `.p12`, etc., just to provide additional protection from users accidentally picking a large or invalid type of file. But I think it's okay not to have this to start out.

src/generated/resources/openapi.yaml
```diff
@@ -183,6 +183,32 @@ paths:
...
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Preferences"
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Preferences"
```
rhauch, line 211: We should add the 400 response here, too, via annotations [like we do in the `ConnectionsResource`](https://github.com/confluentinc/ide-sidecar/blob/0e27f1c878a581e3c4f7491e5a3a6b57500ca87b/src/main/java/io/confluent/idesidecar/restapi/resources/ConnectionsResource.java#L83-L112).

src/main/java/io/confluent/idesidecar/restapi/util/WebClientFactory.java
```diff
@@ -33,16 +39,45 @@ public class WebClientFactory {
...
     return webClient;
   }
+
+  /**
+   * Observer method that is called whenever the {@link Preferences} are updated. Updates the
+   * configuration of the web client according to the passed {@link Preferences} instance.
+   *
+   * @param preferences The new {@link Preferences}
+   */
+  public synchronized void updateWebClientOptions(@Observes PreferencesSpec preferences) {
```
rhauch, line 56: Very nice approach!

rhauch (approved):
Thanks, @flippingbits. I really like this approach, and nice job with the implementation. I have a few comments -- the OpenAPI spec suggestion could be implemented in a quick followup PR if that's easier.

src/main/java/io/confluent/idesidecar/restapi/models/Preferences.java
flippingbits: Good question! I'm all in for validating the input provided to this API endpoint but I'm not sure if validating the file extension will cause more harm than good. I fear that we might reject files holding a valid certificate because they use the wrong file extension. Let's discuss it with the folks working on the extension.

src/generated/resources/openapi.yaml
flippingbits: Great catch! I'll address this in a follow-up PR.

PR #65, opened 2024-09-26
Introduce template option properties `initial_value` and `min_length`

flippingbits:
## Summary of Changes

Update the sidecar to the latest revision of the template registry, which introduced the template option properties `initial_value` and `min_length`:

* Introduce the property `min_length` that specifies the minimum number of characters that users must provide for a template option. By default, `min_length` is set to 0, allowing users to provide empty strings for an option.
* Introduce the property `initial_value`, which can be used by the extension to initialize the form fields in the project wizard.
* Remove the property `default_value` from the option properties and require values for all options.
We must update the VS Code extension to use the new option properties in the PR that updates the extension to a sidecar version holding this change. cc @noeldevelops

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [x] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


noeldevelops (approved):
This looks good @flippingbits 
I made the changes in a draft on the `vscode` side and it will work fine on our form, very simple to update.

I plan to wait until this is merged (after more SSL-related changes if any?) and then do some testing & add this new sidecar version in my branch in `vscode`

PR #66, opened 2024-09-26
Follow-ups for PR #64

flippingbits:
## Summary of Changes

A few smaller follow-ups for https://github.com/confluentinc/ide-sidecar/pull/64:

* Document the `400` response of the endpoint `PUT /gateway/v1/preferences` per @rhauch's suggestion.
* Provide a more meaningful error message in the case of a failed SSL handshake.
* Use "Confluent for VS Code" when referencing the extension on the callback pages.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/resources/OAuthCallbackResource.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/resources/OAuthCallbackResource.java
@@ -39,8 +39,9 @@ public class OAuthCallbackResource {
       .orElse("https://marketplace.visualstudio.com/items?itemName=confluentinc.vscode-confluent");
 
   static final String TLS_HANDSHAKE_ERROR_MESSAGE =
-      "Failed to perform the SSL/TLS handshake. Consider configuring custom certificates in "
-      + "Confluent for VS Code if you are behind a firewall that performs SSL inspection.";
+      "Failed to perform the SSL/TLS handshake. Consider configuring custom certificates in the "
+      + "extension settings of Confluent for VS Code if you are behind a firewall that performs "
+      + "SSL inspection.";
 
   @Inject
   ConnectionStateManager mgr;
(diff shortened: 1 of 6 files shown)


src/main/java/io/confluent/idesidecar/restapi/resources/OAuthCallbackResource.java
```diff
@@ -37,6 +38,10 @@ public class OAuthCallbackResource {
       .getOptionalValue("ide-sidecar.connections.ccloud.oauth.vscode-extension-uri", String.class)
       .orElse("https://marketplace.visualstudio.com/items?itemName=confluentinc.vscode-confluent");
 
+  static final String TLS_HANDSHAKE_ERROR_MESSAGE =
+      "Failed to perform the SSL/TLS handshake. Consider configuring custom certificates in "
+      + "Confluent for VS Code if you are behind a firewall that performs SSL inspection.";
```
rhauch, line 43: The VS Code extension settings appear to be called "SSL/TLS PEM Path", but here we just say "custom certifcates". Should we try to align the wording so that it's easier for users that see this error to find the correct setting?

Perhaps something like this:
```suggestion
      "Failed to perform the SSL/TLS handshake. If you are behind a firewall that performs 
      + "SSL inspection, consider setting the SSL/TLS PEM path in Confluent for VS Code extension settings .";
```

rhauch (commented):
Just one suggestion about the wording of the error.

src/main/java/io/confluent/idesidecar/restapi/resources/OAuthCallbackResource.java
flippingbits: Good question! I wanted to keep it a bit more generic and not mention any specific format because I assumed that we'd add a few more certificate formats in the near term.

What about renaming it to `Failed to perform the SSL/TLS handshake. Consider configuring custom certificates in the extension settings of Confluent for VS Code if you are behind a firewall that performs SSL inspection.` so that users know that they need to head to the **extension settings**?

rhauch: That's a good observation -- we probably don't want to tie it too closely with the particulars of the user settings. I do like your suggested text.

rhauch (approved):
LGTM. Thanks for following up, @flippingbits!

PR #74, opened 2024-10-08
Implement Kafka REST from scratch

rohitsanj:
## TLDR

This PR implements the Topic V3, Cluster V3 Kafka REST APIs from scratch using the Kafka `AdminClient` and swaps the underlying Kafka REST proxy target for confluent-local connections to point to ide-sidecar's own Kafka REST implementation, rather than to confluent-local's kafka-rest server. This will put the ide-sidecar Kafka REST impl in use from day 1 and give us early feedback to iterate.

## Background/Motivation

The ide-sidecar Quarkus application today [exposes a wildcard route](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/resources/ClusterRestProxyResource.java#L33-L37) (what we call the “Kafka REST proxy”) at http://localhost:26636/kafka/v3/clusters/* and uses connection/cluster metadata passed as request headers to determine which Kafka REST server the request must be forwarded to. We want to extend this functionality to allow clients of ide-sidecar to be able to connect to _any_ Kafka cluster, given appropriate configuration context. Based on [this (internal) decision log](https://confluentinc.atlassian.net/wiki/spaces/DTX/pages/3696461455/Decision+Log+Supporting+native+Kafka+protocol+in+ide-sidecar), we decided to implement _from scratch_ a minimal set of Kafka REST endpoints in ide-sidecar, though the major reasons for the current approach are as follows:
* The existing https://github.com/confluentinc/kafka-rest library is designed to be configured against a single Kafka cluster. We've prototyped embedding the library and integrating it into our embedded proxy mechanism, so it's doable.
* The blocker is that the existing https://github.com/confluentinc/kafka-rest library uses Guava, which is not compatible with GraalVM. This makes it impossible to embed this library into the sidecar, as we had originally planned. We will asynchronously work with that project to be compatible with GraalVM, but there may be other challenges besides Guava, so there is no estimated timeframe for completion.
* We are currently using a small subset of the Kafka REST API, so short-term implementing a small portion of the same OpenAPI spec is doable. This PR takes this approach. However, as we need more, we hope to replace the current partial re-implementation with a GraalVM-compatible https://github.com/confluentinc/kafka-rest library.

## Description of code changes

- `kafka-rest.openapi.yaml`: Kafka REST OpenAPI spec taken from confluentinc/kafka-rest that is used to generate API models and resource interfaces. See `pom.xml` for the `openapi-generator-maven-plugin` configuration.
  - The [original spec](https://github.com/confluentinc/kafka-rest/blob/b7b1bba059e878dd28d2dbc5c51d25d3d747bd16/api/v3/openapi.yaml) has been modified to prefix all paths with `/internal/kafka/v3`. Hence, the path [`/clusters/{cluster_id}` in the original spec](https://github.com/confluentinc/kafka-rest/blob/b7b1bba059e878dd28d2dbc5c51d25d3d747bd16/api/v3/openapi.yaml#L83C3-L83C25) is instead [`/internal/kafka/v3/clusters/{cluster_id}` in our modified spec](https://github.com/confluentinc/airlock-ide-sidecar/blob/a7dce67fbbf71a6e4ca972b7587ecade57549096/kafka-rest.openapi.yaml#L42).
- We generate API resource interface classes for the following API groups: TopicV3, ClusterV3, PartitionV3. Out of these, only TopicV3 and ClusterV3 have concrete implementations. PartitionV3 will be implemented in a followup PR to simply the scope of this PR. 
- Note that the OpenAPI generated classes are NOT checked in to the ide-sidecar source code but are instead generated at build time from the `target/generated-sources/openapi/src/gen/java` directory.
- A `Clients` abstract class was extracted out of the `SchemaRegistryClients` class and made to be its parent. The `Clients` class can be parameterized based on the client instance type that needs to be cached by connection id and client id.
  - The new `Clients` class caches client instances by cluster id in a Caffeine cache. 
  - It allows inheritors to define arbitrary cache policies (see [CaffeineSpec](https://github.com/ben-manes/caffeine/wiki/Specification)). 
- A new class `AdminClients extends Clients<org.apache.kafka.clients.admin.AdminClient>`  is used to cache `AdminClient` instances.
  - It sets the Caffeine cache specification to `expiresAfterAccess=5m`, this evicts and closes the `AdminClient` client instances after 5 minutes of inactivity.
- The exception mappers related to Kafka REST exceptions construct response models using the kafka-rest OpenAPI generated `io.confluent.idesidecar.restapi.kafkarest.model.Error` model, as opposed to using ide-sidecar's own `io.confluent.idesidecar.restapi.exceptions.Failure.Error` model. This is intentional in order to adhere to the kafka-rest OpenAPI spec as much as possible.
- [Major change] The Kafka REST Proxy target for `LOCAL` (confluent local) connections was changed from the kafka-rest server (served at `http://localhost:8082/v3` in the confluent-local docker container), to ide-sidecar's own kafka-rest implementation served at `http://localhost:26636/internal/kafka/v3`.
  - Given that the PartitionAPI has not yet been implemented in ide-sidecar's Kafka REST, listing partitions for a confluent-local kafka topic will NOT work temporarily.
  - **Motivation behind this change**: Swapping the confluent-local proxy target to now point to ide-sidecar's Kafka REST will give us early feedback to harden/iterate on the implementation.
- A few tests that tested prior implementation path for confluent-local Kafka REST proxy have been removed since they are no longer compatible with current implementation. Local connectivity has been sufficiently tested in other integration tests that actually spin up a confluent-local docker container. 

## Design considerations

- Clients of ide-sidecar must only be aware of the `http://localhost:26636/kafka/v3` Kafka REST proxy routes. The `http://localhost:26636/internal/kafka/v3` endpoints serve as an implementation detail for talking to clusters using the Kafka-native protocol.

## Followup items

- PartitionV3 API implementation (see #77)
- Include connection-specific resource names from the internal Kafka REST responses
- From Stefan's [comment](https://github.com/confluentinc/ide-sidecar/pull/74#discussion_r1794797091), set `client.id` to something like `Quarkus for VS Code sidecar ${quarkus.application.version}`


rhauch:
Thanks, @rohitsanj. Please correct me if I'm wrong, but after this change we'll have two Kafka REST endpoints:
1. http://localhost:26636/internal/kafka/v3 -- This is the "embedded" Kafka REST API that will always translate requests into Kafka-native protocol requests submitted directly against the external designated Kafka cluster, and translate the Kafka-native responses back to the RESTful representation. This "embedded" Kafka REST API is not meant to be called by the extension.
2. http://localhost:26636/kafka/v3 -- This is the existing Kafka REST API that the extension uses to talk to an external Kafka REST API server that is running next to a Kafka cluster. With this change, this proxy Kafka REST API will forward requests _either_ to the external Kafka REST server, _or_ forward (or-redirect?) requests to the internal "embedded" Kafka REST API that will forward requests (over the Kafka-native protocol) to the external Kafka cluster.

(The endpoints for both APIs, like nearly all the sidecar APIs) both require the session token be passed as the authz header.)

Can you confirm that this approach was chosen because it makes implementing the `http://localhost:26636/kafka/v3` endpoints straightforward, since they always forward to an external or internal Kafka REST server?

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
```diff
@@ -0,0 +1,105 @@
...
+        .forConnection(connectionId)
+        // Accessing a ConcurrentHashMap _may_ be a blocking operation
+        .kafkaClusters
+        .values()
+        .stream()
+        .findFirst()
+        .orElseThrow(
+            () -> new AdminClientInstantiationException(
+                "No Kafka clusters found for connection " + connectionId
+            )
```
rhauch, line 81: Do we want to throw ClusterNotFoundException here, which is a bit more accurate? Or do we need this `AdminClientInstantiationException` type in particular?

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
```diff
@@ -0,0 +1,105 @@
...
+        ).spec();
+
+    if (cluster.bootstrapServers() == null) {
+      throw new AdminClientInstantiationException(
+          "No bootstrap servers found for Kafka cluster " + cluster.id()
+      );
+    }
```
rhauch, line 88: Is it really possible that the `ClusterCache` has a `KafkaCluster` object with a null `bootstrapServers()` value? Seems like we should have caught this earlier in the `ClusterCache` itself.

If it's not possible, then perhaps we should update the JavaDoc for `KafkaCluster.bootstrapServers()` that says it never returns null, or we annotate `KafkaCluster.bootstrapServers` with `io.smallrye.common.constraint.NotNull`

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
```diff
@@ -0,0 +1,105 @@
...
+
+    var props = new Properties();
+    props.put("bootstrap.servers", cluster.bootstrapServers());
+    props.putAll(adminClientDefaults);
+    return props;
```
rhauch, line 93: This code is duplicated. Should we move this to a helper method?

src/main/java/io/confluent/idesidecar/restapi/cache/Clients.java
```diff
@@ -0,0 +1,129 @@
...
+   */
+  private final Map<String, Cache<String, T>> clientsByIdByConnections =
+      new ConcurrentHashMap<>();
+
+  RemovalListener<String, T> removalListener = (key, value, cause) -> {
+    try {
+      if (value != null) {
+        Log.debugf("Closing client %s", value);
+        value.close();
+      }
```
rhauch, line 33: What do you think about making this function a member method, like:
```
   void handleRemoval(String key, T value, RemovalCause cause) { ... }
```
Then `createCache()` would reference the method `this::handleRemoval`. (This should work since `createCache()` is not called during the `AdminClient` constructor.)

src/main/java/io/confluent/idesidecar/restapi/kafkarest/controllers/RelationshipUtil.java
```diff
@@ -0,0 +1,118 @@
...
+    );
+  }
+
+  public static Crn getTopicCrn(String clusterId, String topicName) {
+    return new Crn(
+        CRN_AUTHORITY,
+        Crn.newElements("kafka", clusterId, "topic", topicName),
+        false
+    );
+  }
```
rhauch, line 117: Should these be in `Crn` like the other methods? Yes, these are not CCloud related, but that's okay.

src/main/java/io/confluent/idesidecar/restapi/kafkarest/controllers/ClusterManagerImpl.java
```diff
@@ -0,0 +1,144 @@
...
+        );
+  }
+
+  private ClusterData fromClusterId(ClusterDescribe cluster) {
+    return ClusterData
+        .builder()
+        .kind("KafkaCluster")
+        .metadata(ResourceMetadata
+            .builder()
+            .self(getClusterCrn(cluster.id()).toString())
```
rhauch, line 103: Should we include CRN if this is not CCloud? If it is CCloud, then can we not use the existing `Crn` factory methods?

src/main/java/io/confluent/idesidecar/restapi/kafkarest/controllers/RelationshipUtil.java
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Utility class for generating {@code related} links for Kafka resources.
+ */
+public final class RelationshipUtil {
+
+  private RelationshipUtil() {
+  }
+
+  private static final String CRN_AUTHORITY = "";
```
rhauch, line 15: Why is the authority blank here? That's not a valid CRN (or even resource name)

src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/strategy/ConfluentLocalKafkaClusterStrategy.java
```diff
@@ -1,45 +1,64 @@
...
+        .add(CONNECTION_ID_HEADER, context.getConnectionId());
   }
 
-
+  /**
+   * Route the request back to ourselves at the /internal/kafka path.
+   * Context: We used to send this proxy request to the Kafka REST server running in the
+   * confluent-local container, but now we route the request to our own implementation of the
+   * Kafka REST API, served at the /internal/kafka path. This was done to get early feedback
+   * on our in-house implementation of the Kafka REST API.
```
rhauch, line 39: Why just "early feedback"? Is this not the longer term direction?

src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/strategy/ConfluentLocalKafkaClusterStrategy.java
```diff
@@ -1,45 +1,64 @@
...
-        "http://%s".formatted(confluentLocalKafkaRestHostname),
-        // Since the REST API running in Confluent Local Kafka is configured to run on /v3,
-        // we add the /kafka to the sidecar URI before replacing the cluster URI
-        // with the sidecar URI.
-        uriUtil.combine(sidecarHost, "/kafka")
-    );
+  public String processProxyResponse(String proxyResponseBody) {
+    return super
+        .processProxyResponse(proxyResponseBody)
+        .replaceAll("/internal/kafka", "/kafka");
```
rhauch, line 62: This is a lot less specific than including `https://<host>`. Should we also include that, just because we're doing a blind string-based replace-all on the response?

rhauch (commented):
Thanks for this excellent start, @rohitsanj.

rohitsanj:
> Can you confirm that this approach was chosen because it makes implementing the http://localhost:26636/kafka/v3 endpoints straightforward, since they always forward to an external or internal Kafka REST server?

(Ah, I think I meant to mention the reasoning in the description but missed it.)

Yes, your reasoning about the endpoint implementation is definitely valid, however I think there's greater benefit in terms of abstracting away from the client as to whether a particular connection type requires talking to an external or internal Kafka REST endpoint. I will update the PR description to add this detail.

I will also further reinforce the abstraction by removing the `/internal/kafka/v3` endpoints from the generated OpenAPI specs.

> The endpoints for both APIs, like nearly all the sidecar APIs) both require the session token be passed as the authz header.

Thanks for mentioning this. I will update the logic to add the authz header as well when fwding the request to internal Kafka REST.

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /**
+   * Get AdminClient from the first available Kafka cluster for the connection. We implicitly assume
+   * that there will be only one Kafka cluster per connection.
+   * @param connectionId The connection ID
+   * @return The AdminClient configuration
+   */
+  public Properties getAdminClientConfig(String connectionId) {
```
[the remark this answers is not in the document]
rohitsanj: I went back and forth on doing the `AdminClient.create` call either in this class or elsewhere but kept the logic to construct the config here in the App-scoped `AdminClients` bean. It must be a relic from that, I will change these to private.

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
```diff
@@ -0,0 +1,105 @@
...
+          () -> AdminClient.create(getAdminClientConfig(connectionId))
+      );
+    }
+  }
+
+  public Properties getAdminClientConfig(String connectionId, String clusterId) {
+    var props = new Properties();
+    var cluster = cache.getKafkaCluster(connectionId, clusterId);
+    props.put("bootstrap.servers", cluster.bootstrapServers());
+    props.putAll(adminClientDefaults);
```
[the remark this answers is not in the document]
rohitsanj: You're right. I'll come up with a better name and move it above.

src/main/java/io/confluent/idesidecar/restapi/cache/Clients.java
```diff
@@ -0,0 +1,129 @@
...
+  public T getClient(
+      String connectionId,
+      String clientId,
+      Supplier<T> factory
+  ) {
+    return clientsForConnection(connectionId).asMap().computeIfAbsent(
+        clientId,
+        k -> factory.get()
+    );
+  }
```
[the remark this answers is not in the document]
rohitsanj: Hm, we do only want to use the `getAdminClient` on `AdminClients`. 

I see it like this: if the client (of the `AdminClients` class has additional context not already provided by the `getAdminClient` method, then they may choose to use the `getClient` method and provide the supplier function. I guess I'll think about this some more to make the interface cleaner/clearer.

src/main/java/io/confluent/idesidecar/restapi/kafkarest/controllers/RelationshipUtil.java
rohitsanj: I inferred it from examining an API response from confluent-local's kafka-rest server. Taking a closer look at confluentinc/kafka-rest, I see that they've defaulted the Crn authority to an empty string as well.

https://github.com/confluentinc/kafka-rest/blob/b7b1bba059e878dd28d2dbc5c51d25d3d747bd16/kafka-rest/src/main/java/io/confluent/kafkarest/KafkaRestConfig.java#L413-L415

src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/strategy/ConfluentLocalKafkaClusterStrategy.java
rohitsanj: Yes, it is the longer term direction. By "early feedback", I meant that if we simply add the Kafka REST implementation to ide-sidecar without having it be the backbone of an existing feature set (confluent local), then we'd only discover any latent issues/bugs (not covered by integration tests) once we actually _do_ start using it end-to-end from the VS Code extension.

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
rohitsanj: I'll take a closer look and see if we need the additional check here.

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
rohitsanj: I think we need the `AdminClientInstantiationException` specifically here because we will only get to this code path when listing clusters (`GET /internal/kafka/v3/clusters`) on a connection that has no Kafka clusters associated, leading to a `400` status code (open to making this more accurate) -- see [exception mapper](https://github.com/confluentinc/ide-sidecar/blob/7c04ed8469dce896fe5d75f38ab94ce852370d57/src/main/java/io/confluent/idesidecar/restapi/exceptions/ExceptionMappers.java#L228-L240). However, using `ClusterNotFoundException` results in a 404 which feels inaccurate.

Considering that the internal endpoints are NOT meant to be _directly_ invoked by clients of ide-sidear, I figured simply handling the error scenario and bubbling up mostly appropriate error status codes/messages is sufficient.

src/main/java/io/confluent/idesidecar/restapi/cache/Clients.java
```diff
@@ -0,0 +1,129 @@
...
+
+  int clientCount() {
+    return clientsByIdByConnections
+        .values()
+        .stream()
+        .map(Cache::asMap)
+        .map(Map::size)
+        .mapToInt(Integer::intValue)
+        .sum();
+  }
```
flippingbits, line 71: Does this method return a primitive type because of the limitations of `IntStream.sum()`?

src/main/java/io/confluent/idesidecar/restapi/kafkarest/controllers/ClusterManager.java
```diff
@@ -0,0 +1,15 @@
+package io.confluent.idesidecar.restapi.kafkarest.controllers;
```
flippingbits, line 1: Is this package meant to hold controller classes per the MVC pattern?

src/main/java/io/confluent/idesidecar/restapi/kafkarest/controllers/TopicManagerImpl.java
```diff
@@ -0,0 +1,156 @@
...
+            Optional.ofNullable(includeAuthorizedOperations).orElse(false)
+        );
+    return clusterManager.getKafkaCluster(clusterId).chain(ignored -> uniStage(
+        adminClients.getAdminClient(connectionId.get(), clusterId)
+            .describeTopics(List.of(topicName), describeTopicsOptions)
+            .allTopicNames()
+            .toCompletionStage()
+    )
+        .map(topicDescriptions -> topicDescriptions.values().iterator().next())
+        .onItem().transform(topicDescription -> fromTopicDescription(clusterId, topicDescription)));
```
flippingbits, line 97: nit: This is hard to read. What about something like:

```suggestion
    return clusterManager
        .getKafkaCluster(clusterId)
        .chain(ignored ->
            uniStage(
              adminClients
                  .getAdminClient(connectionId.get(), clusterId)
                  .describeTopics(List.of(topicName), describeTopicsOptions)
                  .allTopicNames()
                  .toCompletionStage()
            )
            .map(topicDescriptions -> topicDescriptions.values().iterator().next())
            .onItem()
            .transform(topicDescription -> fromTopicDescription(clusterId, topicDescription))
        );
```

src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/strategy/ConfluentLocalKafkaClusterStrategy.java
flippingbits: Do we need to perform the replacement? I mean, technically, the `/internal/kafka/...` routes are valid and working.

rohitsanj: I think we should perform the replacement given that we want to (at least initially) hide the `/internal/kafka` routes from ide-sidecar clients.

src/main/java/io/confluent/idesidecar/restapi/cache/Clients.java
rohitsanj: Hmm, I'm not sure. FWIW I didn't author this function per se, it was moved over from `SchemaRegistryClients`. It's only used in tests.

Also, TIL about `IntStream`.

src/main/java/io/confluent/idesidecar/restapi/kafkarest/controllers/ClusterManager.java
rohitsanj: No, I don't think it's meant to represent MVC. Perhaps just calling the package `services` would be more appropriate? WDYT?

src/main/java/io/confluent/idesidecar/restapi/kafkarest/controllers/ClusterManagerImpl.java
rohitsanj: No, I've made `resource_name` null. Also, `self` here should've been `RelationshipUtil.forCluster(clusterId)` and not a CRN. I've fixed that.

src/main/java/io/confluent/idesidecar/restapi/kafkarest/controllers/RelationshipUtil.java
rohitsanj: Removed these methods. We decided to tackle constructing CRNs in a follow up.

src/main/java/io/confluent/idesidecar/restapi/kafkarest/controllers/RelationshipUtil.java
rohitsanj: Removed this.

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
rohitsanj: I've removed the possibly redundant check here and annotated the `KafkaCluster.bootstrapServers` with `io.smallrye.common.constraint.NotNull`.

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
rohitsanj: Going to let this be `AdminClientInstantiationException` unless you think otherwise, given my reasoning above.

src/main/java/io/confluent/idesidecar/restapi/kafkarest/filters/ConnectionIdHeaderFilter.java
```diff
@@ -0,0 +1,64 @@
+package io.confluent.idesidecar.restapi.kafkarest.filters;
```
rhauch, line 1: Could this move to `io.confluent.idesidecar.restapi.filters`?

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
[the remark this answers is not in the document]
rhauch: Yeah, we can always expand visibility in the future, but it's better to start with a smaller and more succinct public API. Definitely don't remove the method(s), since they are very useful.

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
rhauch: > I think we need the AdminClientInstantiationException specifically here because we will only get to this code path when listing clusters (GET /internal/kafka/v3/clusters) on a connection that has no Kafka clusters associated, leading to a 400 status code (open to making this more accurate) -- see [exception mapper](https://github.com/confluentinc/ide-sidecar/blob/7c04ed8469dce896fe5d75f38ab94ce852370d57/src/main/java/io/confluent/idesidecar/restapi/exceptions/ExceptionMappers.java#L228-L240). However, using ClusterNotFoundException results in a 404 which feels inaccurate.

Hmm, seems like listing clusters (GET /internal/kafka/v3/clusters) on a connection that has no Kafka clusters associated would return an empty list, not fail. No?

src/main/java/io/confluent/idesidecar/restapi/cache/Clients.java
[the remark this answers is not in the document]
rhauch: I guess it's weird that the `AdminClients` class has these methods:
* `getClient(String, String, Supplier<AdminClient>): AdminClient`
* `getAdminClient(String, String): AdminClient`
* `getAdminClient(String): AdminClient`

Also, the `SchemaRegistryClients` class has only this inherited method:
* `getClient(String, String, Supplier<SchemaRegistryClient>)`

Whether or not the `AdminClients.getClient(String, String, Supplier<AdminClient>)` method is public, perhaps the methods should all be `getClient(...)` just with different parameters. They all are doing the same thing, and method overloading works really nicely here.

src/main/java/io/confluent/idesidecar/restapi/cache/Clients.java
```diff
@@ -0,0 +1,129 @@
...
+   * Get the Caffeine spec to use for the caches. By default, this will be an empty spec.
+   * See the
+   * <a href="https://github.com/ben-manes/caffeine/wiki/Specification">Caffeine Spec</a>
+   * for more information on the format. Inherited classes may override this method
+   * and return a different spec.
+   * @return the Caffeine spec
+   */
+  protected CaffeineSpec getCaffeineSpec() {
+    return CaffeineSpec.parse("");
+  }
```
rhauch, line 95: The implementation of `getCaffeineSpec()` has no parameters and there's no `Clients` state, which means in practice the specs will always be the same. Is there a benefit to defining the spec with a method, versus passing the spec into the `Clients` constructor as an immutable spec for use in all embedded caches?

src/main/java/io/confluent/idesidecar/restapi/kafkarest/controllers/ClusterManager.java
rhauch: Did you consider not having any subpackages under `kafkarest`? As mentioned my other comment, the filter could very easily go to `...restapi.filters`, and the `AdminClientInstantiationException` to `...restapi.exceptions`. That leaves just the `impl` and `controllers` packages for a total of eight classes.

rohitsanj: I think that's fine. I'll make the eight classes be top-level under `kafkarest` and move the other two out.

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
rohitsanj: You.. would be right, thanks! I didn't see it that way initially, but makes total sense.

src/main/java/io/confluent/idesidecar/restapi/cache/Clients.java
rohitsanj: Makes sense, thanks! I'll incorporate that.

src/main/java/io/confluent/idesidecar/restapi/cache/Clients.java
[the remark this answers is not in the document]
rohitsanj: It is weird. Overloading does work really nicely, thanks.

rohitsanj:
I expect a few integration tests to fail in CI due to the following error. 

<details>
<summary>Caffeine error</summary>
```
2024-10-09 14:11:48,897 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (vert.x-eventloop-thread-10) HTTP Request to /internal/kafka/v3/clusters/oh-sxaDRTcyAr6pFRbXyzA/topics failed, error id: b4f8f46f-54f1-4752-9461-de562b4fc7b3-13: java.lang.IllegalStateException: SSLA
	at com.github.benmanes.caffeine.cache.LocalCacheFactory.newFactory(LocalCacheFactory.java:114)
	at java.base@21.0.2/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1708)
	at com.github.benmanes.caffeine.cache.LocalCacheFactory.loadFactory(LocalCacheFactory.java:97)
	at com.github.benmanes.caffeine.cache.LocalCacheFactory.newBoundedLocalCache(LocalCacheFactory.java:46)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache$BoundedLocalManualCache.<init>(BoundedLocalCache.java:3953)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache$BoundedLocalManualCache.<init>(BoundedLocalCache.java:3949)
	at com.github.benmanes.caffeine.cache.Caffeine.build(Caffeine.java:1048)
	at io.confluent.idesidecar.restapi.cache.Clients.createCache(Clients.java:102)
	at io.confluent.idesidecar.restapi.cache.Clients.lambda$clientsForConnection$1(Clients.java:49)
	at java.base@21.0.2/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1708)
	at io.confluent.idesidecar.restapi.cache.Clients.clientsForConnection(Clients.java:49)
	at io.confluent.idesidecar.restapi.cache.Clients.getClient(Clients.java:42)
	at io.confluent.idesidecar.restapi.cache.AdminClients.getAdminClient(AdminClients.java:44)
	at io.confluent.idesidecar.restapi.cache.AdminClients_ClientProxy.getAdminClient(Unknown Source)
	at io.confluent.idesidecar.restapi.kafkarest.services.ClusterManagerImpl.lambda$describeCluster$3(ClusterManagerImpl.java:84)
	at io.confluent.idesidecar.restapi.kafkarest.services.ClusterManagerImpl.lambda$describeCluster$4(ClusterManagerImpl.java:86)
	at io.smallrye.context.impl.wrappers.SlowContextualFunction.apply(SlowContextualFunction.java:21)
	at io.smallrye.mutiny.operators.uni.UniOnItemTransform$UniOnItemTransformProcessor.onItem(UniOnItemTransform.java:36)
	at io.smallrye.mutiny.operators.uni.UniOperatorProcessor.onItem(UniOperatorProcessor.java:47)
	at io.smallrye.mutiny.operators.uni.builders.UniCreateFromKnownItem$KnownItemSubscription.forward(UniCreateFromKnownItem.java:38)
	at io.smallrye.mutiny.operators.uni.builders.UniCreateFromKnownItem.subscribe(UniCreateFromKnownItem.java:23)
	at io.smallrye.mutiny.operators.AbstractUni.subscribe(AbstractUni.java:36)
	at io.smallrye.mutiny.operators.uni.UniRunSubscribeOn.lambda$subscribe$0(UniRunSubscribeOn.java:27)
	at io.quarkus.vertx.core.runtime.VertxCoreRecorder$14.runWith(VertxCoreRecorder.java:635)
	at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2516)
	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2495)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1521)
	at org.jboss.threads.DelegatingRunnable.run(DelegatingRunnable.java:11)
	at org.jboss.threads.ThreadLocalResettingRunnable.run(ThreadLocalResettingRunnable.java:11)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base@21.0.2/java.lang.Thread.runWith(Thread.java:1596)
	at java.base@21.0.2/java.lang.Thread.run(Thread.java:1583)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine(PlatformThreads.java:833)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.posix.thread.PosixPlatformThreads.pthreadStartRoutine(PosixPlatformThreads.java:211)
Caused by: java.lang.ClassNotFoundException: com.github.benmanes.caffeine.cache.SSLA
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.hub.ClassForNameSupport.forName(ClassForNameSupport.java:122)
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.hub.ClassForNameSupport.forName(ClassForNameSupport.java:86)
	at java.base@21.0.2/java.lang.Class.forName(DynamicHub.java:1356)
	at java.base@21.0.2/java.lang.Class.forName(DynamicHub.java:1345)
	at java.base@21.0.2/java.lang.invoke.MethodHandles$Lookup.findClass(MethodHandles.java:2869)
	at com.github.benmanes.caffeine.cache.LocalCacheFactory.newFactory(LocalCacheFactory.java:104)
	... 33 more
```

</details>

rohitsanj:
Alright, @rhauch @flippingbits -- thanks a lot for your reviews/feedback! I've incorporated/addressed all (if not most) of it and I think the PR is in a much better shape now. Please have another look, thanks.

rohitsanj:
I made some slight tweaks/fixes identified when testing this branch's sidecar executable against a debug instance of confluentinc/vscode:

1. Set `content-length` only if `transfer-encoding` is not set. From [this section in RFC 7230: HTTP 1.1 Message Syntax and Routing](https://www.rfc-editor.org/rfc/rfc7230#section-3.3.2): _  A sender MUST NOT send a Content-Length header field in any message
   that contains a Transfer-Encoding header field._.
2. Map exceptions of type `org.apache.kafka.common.errors.ApiException` raised by `AdminClient` operations, to status code 500, with exception message passed through as is.
3. Added support for the `includeAuthorizedOperations` query parameter (notice the `camelCase`) to the list topics endpoint. This is used by the VS Code extension to determine the scope of actions a user may perform on a topic. (See [this comment](https://github.com/confluentinc/ide-sidecar/pull/74/files#diff-d7a2124f9604ec0ef290d171941040875ea004f82c75cd94d079aa366ccbc6c4R1595-R1599) in the `kafka-rest.openapi.yml` file for more context).

I also ended up reverting changes made to `ClusterStrategy` since I felt the refactor did not bring much benefit.

kafka-rest.openapi.yaml
```diff
@@ -0,0 +1,4545 @@
...
+      in: query
+      required: false
+      schema:
+        type: boolean
+
+    # NOTE: Kafka-rest's OpenAPI spec has a known undocumented inconsistency where
+    #       the list topics operation expects the includeAuthorizedOperations query parameter
+    #       whereas the get topic operation expects the include_authorized_operations query parameter.
+    #       We declare both here in our version of the spec to at least ensure that it is at least documented.
+    #       The inconsistency has been set in stone for a while and is unlikely to change in the near future.
```
rohitsanj, line 1599: @jlrobins: I believe this comment here about Kafka REST's API spec inconsistency might strike a chord with you.

src/main/resources/application.yml
```diff
@@ -119,6 +119,11 @@ ide-sidecar:
     enabled: true
     # How often to check?
     interval-seconds: 60
+  # Configuration set by the sidecar when instantiating the Kafka Admin client
+  # (We may choose to allow connection-specific overrides in the future)
+  admin-client-configs:
+    "default.api.timeout.ms": 3000
+    "request.timeout.ms": 3000
```
flippingbits, line 126: @rohitsanj @rhauch What are your thoughts on including a sidecar-specific `client.id`, like `Confluent ide-sidecar ${quarkus.application.version}` (`${quarkus.application.version}` should evaluate to the current release) or similar, in these default configs so that we/users can attribute traffic to the IDE sidecar? This might be especially useful when using the `AdminClient` to interact with remote clusters (CP or CCloud).

rhauch: Yeah, that's a good suggestion. Do we want to just say "sidecar" rather than "VS Code"? Maybe:
```
Confluent for VS Code sidecar ${quarkus.application.version}
```

rohitsanj:
> Did you manually create this file or did you use the [GraalVM tracing agent](https://www.graalvm.org/latest/reference-manual/native-image/metadata/AutomaticMetadataCollection/)?

@flippingbits, I manually created the file based on [this Quarkus guide](https://quarkus.io/guides/writing-native-applications-tips#using-a-configuration-file-2). I considered using the [GraalVM reachability metadata repository](https://www.graalvm.org/latest/reference-manual/native-image/guides/use-reachability-metadata-repository-maven/) since `com.github.ben-manes.caffeine:caffeine` has community-contributed metadata (see exhaustive list [here](https://www.graalvm.org/native-image/libraries-and-frameworks/)).

src/main/resources/application.yml
rohitsanj: That is indeed a 💯 suggestion, thanks Stefan! I would say we do this in a follow up given the scope and complexity of this PR. I imagine there will be some dynamic string formatting involved to populate the sidecar version in the id string.

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
```diff
@@ -0,0 +1,61 @@
...
+    // Set AdminClient configs provided by the sidecar
+    props.putAll(adminClientSidecarConfigs);
+    props.put("bootstrap.servers", cluster.bootstrapServers());
+    return props;
+  }
+
+  @Override
+  void onConnectionUpdated(@ObservesAsync @Lifecycle.Updated ConnectionState connection) {
+    clearClients(connection.getId());
+  }
```
rhauch, line 60: I think we need to do this for SR clients, too, to handle cases where the SR hostname+port have been changed. Maybe we can just move this into the base class?

src/main/java/io/confluent/idesidecar/restapi/cache/Clients.java
```diff
@@ -0,0 +1,134 @@
...
+   */
+  private final Map<String, Cache<String, T>> clientsByIdByConnections =
+      new ConcurrentHashMap<>();
+
+  protected Clients(CaffeineSpec caffeineSpec) {
+    this.caffeineSpec = caffeineSpec;
+  }
+
+  protected Clients() {
+    this.caffeineSpec = CaffeineSpec.parse("");
```
rhauch, line 41: Maybe it doesn't matter in this simple case, but it's often helpful to have one constructor that is used by others, so that initialization logic is centralized.
```suggestion
    this(CaffeineSpec.parse(""));
```

src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
```diff
@@ -209,7 +225,7 @@ protected Clusters createOrGetConnectionClusters(
     );
   }
 
-  record ClusterInfo<SpecT extends Cluster>(
+  public record ClusterInfo<SpecT extends Cluster>(
```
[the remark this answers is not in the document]
rohitsanj: Not sure why IntelliJ thinks "Class 'Clusters' is exposed outside its defined visibility scope" for the `createOrGetConnectionClusters` method. See screenshot:

Screenshot 2024-10-10 at 11 35 11 AM

Anyway, this doesn't _need_ to be made public, I'll change it back. Same for `ClusterInfo`.

src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
```diff
@@ -219,7 +235,7 @@ record ClusterInfo<SpecT extends Cluster>(
     public static final String BLANK_PATH = "";
   }
 
-  class Clusters {
+  public class Clusters {
```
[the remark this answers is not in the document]
rohitsanj: Not sure why IntelliJ thinks "Class 'Clusters' is exposed outside its defined visibility scope" for the `createOrGetConnectionClusters` method. See screenshot:

Screenshot 2024-10-10 at 11 35 11 AM

Anyway, this doesn't _need_ to be made public, I'll change it back. Same for `ClusterInfo`.

src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
```diff
@@ -91,6 +92,21 @@ public KafkaCluster getKafkaCluster(String connectionId, String clusterId) {
     return forConnection(connectionId).getKafkaCluster(clusterId);
   }
 
+  /**
+   * Find the first Kafka cluster accessible over the specified connection. This is useful when
+   * it is known that there is only one Kafka cluster per connection.
+   * @param connectionId the ID of the connection
+   * @return the info for the first Kafka cluster, or null if none found
+   */
+  public Optional<ClusterInfo<KafkaCluster>> getKafkaClusterForConnection(String connectionId) {
```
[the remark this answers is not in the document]
rohitsanj: I'll change this to return `Optional<KafkaCluster>` instead like the other methods, thanks.

src/main/java/io/confluent/idesidecar/restapi/cache/Clients.java
rohitsanj: Thanks! Will keep in mind.

rohitsanj:
@rhauch May I get you to please take another look? Thanks 🙇

rhauch (approved):
Nice work, @rohitsanj. Let's get @flippingbits approval, too.

Can you also clarify what the impact will be on the UX with just this merged, but before #77 is merged? I'd rather us not get stuck with some functionality broken for more than a day or so.

rohitsanj:
Thank you!

> Can you also clarify what the impact will be on the UX with just this merged, but before https://github.com/confluentinc/ide-sidecar/pull/77 is merged? I'd rather us not get stuck with some functionality broken for more than a day or so.

The partitions dropdown in the message viewer tabs for confluent local topics will have "All partitions" as the only option with no ability to filter down on topic partitions. The UX/functionality remains intact for CCloud topic message viewer tabs.

Screenshot 2024-10-10 at 3 02 37 PM

src/main/resources/application.yml
flippingbits: > Yeah, that's a good suggestion. Do we want to just say "sidecar" rather than "VS Code"? Maybe:
> Confluent for VS Code sidecar ${quarkus.application.version}

I didn't suggest "VS Code" but your proposal works for me 👍

flippingbits: > That is indeed a 💯 suggestion, thanks Stefan! I would say we do this in a follow up given the scope and complexity of this PR. I imagine there will be some dynamic string formatting involved to populate the sidecar version in the id string.

I feel this could be a one-liner in the `application.yml` but we can also address it in a follow-up PR if you like.

flippingbits (approved):

PR #89, opened 2024-10-16

ConfluentSemaphore:


rhauch (approved):
Interesting that the generated notice files changed the Apache Commons Logging copyright dates from `2003-2014` to `2001-2024`. The ending year change makes sense, but the beginning year went backwards. ¯\_(ツ)_/¯

PR #91, opened 2024-10-17
Load certificates from Windows-ROOT cert store

flippingbits:
## Summary of Changes

On Windows, load certificates from the cert store `WINDOWS-ROOT` in addition to `WINDOWS-MY`.

Also, update the level of a few log statements so that we can better troubleshoot situations where users have difficulty loading certificates.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/util/WebClientFactory.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/util/WebClientFactory.java
@@ -23,6 +23,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateEncodingException;
 import java.security.cert.CertificateException;
+import java.util.List;
 import javax.net.ssl.TrustManagerFactory;
 import javax.net.ssl.X509TrustManager;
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/idesidecar/restapi/util/WebClientFactory.java
```diff
@@ -163,14 +166,14 @@ void addCertsFromBuiltInTrustStore(PemTrustOptions pemTrustOptions) {
...
-      Log.debug("Loading certificates from system key store.");
+      Log.infof("Loading certificates from system key store %s.", keyStoreType);
       var keyStore = KeyStore.getInstance(keyStoreType);
       keyStore.load(null, null);
       var it = keyStore.aliases().asIterator();
       while (it.hasNext()) {
         var certAlias = it.next();
         var cert = keyStore.getCertificate(certAlias);
-        Log.debugf(
+        Log.infof(
```
rhauch, line 176: How many log lines will this generate, typically? A few or many?

rhauch (commented):
I think this change makes sense and I'm happy to approve, pending successful testing on Windows. I do have one question about the number of `Adding certificate ...` info log messages we might typically see. A handful or even a dozen is fine, but more than a few dozen might be problematic.

flippingbits:
> I do have one question about the number of Adding certificate ... info log messages we might typically see. A handful or even a dozen is fine, but more than a few dozen might be problematic.

Good point -- thanks for raising the concern! Note that these log messages are needed to debug loading certs from the system cert stores on Windows. We can probably revert the changed log level in the next extension release or the release after the next one, once loading certs on Windows is more stable. The messages are only logged when initializing the `WebClient`, which happens once at startup and every time the user updates the preferences. I'd expect a few dozen certs/log messages but it depends on the number of certs maintained in the system cert store.

rhauch (approved):
LGTM. This still adds a few more info-level log messages, and only outputs the _number_ of certificates added.

shouples (approved):
Works great on Windows!

PR #99, opened 2024-10-21
Proxy HTTP request headers from VS Code

### LINKED ISSUE #92: From-vscode HTTP request headers should be proxied
The proxying code for either kafka rest or schema registry needs to proxy not only the request body, but also the request headers. For example, creating schemas via the [SR register route](https://docs.confluent.io/cloud/current/api.html#tag/Subjects-(v1)/operation/register) only [works consistently](https://github.com/confluentinc/vscode/issues/388) if the content-type is set to "application/json" (read all the way through to the bottom of the issue).

I suggest that the processors work together to ultimately make a new MultiMap for the proxy request headers through a process not unlike:

  * Start with copy of the request to-be-proxied headers
  * Strip out the headers meant only for sidecar to consume (`x-cluster-id` , `x-connection-id`)
  * Add in any new headers needed (like sidecar-managed authorization auth token stuff)

So that efforts made on vscode-side to add in extra headers if / when necessary are honored.

While here, it'd also be AWESOME if there were a quarkus setting we could set when sidecar is launched that would put the proxy code into debugging mode, ala `curl -v` so that we can see the request URL, headers, and body plus then the response headers + body explicitly logged out.

rohitsanj:
## Summary of Changes

Include all headers passed from VS Code when proxying requests to a target Kafka/SR cluster, except for sidecar-specific headers (specified as config in `application.yml`).

Resolves #92 

## Pull request checklist

- Tests:
    - [ ] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/strategy/ClusterStrategy.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/strategy/ClusterStrategy.java
@@ -25,11 +25,8 @@ public abstract class ClusterStrategy {
 
   public MultiMap constructProxyHeaders(ClusterProxyContext context) {
     var headers = HttpHeaders.headers();
-    context.getRequestHeaders().forEach(header -> {
-      if (!httpHeaderExclusions.contains(header.getKey())) {
-        headers.add(header.getKey(), header.getValue());
-      }
-    });
+    headers.addAll(context.getRequestHeaders());
+    httpHeaderExclusions.forEach(headers::remove);
     return headers;
   }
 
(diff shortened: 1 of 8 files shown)


rhauch (dismissed):
LGTM

jlrobins:
Worked with Rohit and fixed some issues found in clicktesting. We wrote https://github.com/confluentinc/ide-sidecar/issues/102 in the process thereof.

But this works better than before now. Can now POST to create a schema consistently now that content-type header is making it through to ccloud.

jlrobins (approved):

PR #104, opened 2024-10-22
[v0.49.x] Fix: Add @Blocking annotation to ConnectionsResource

rohitsanj:
## Summary of Changes

The VS Code extension issues get connection requests for every connection at a period of 10 seconds. These requests may block due to lock contention or DNS issues and hence we should run them on the Quarkus worker thread pool instead of the event loop threads so other requests made to the sidecar are not affected.

## Background/Context

A user reported that they were unable to view topics in a message. From their logs, we gleaned that their host machine was not able to resolve DNS for the CCloud server (reason still unknown) when trying to check authentication status. This request then blocked one of the event loop threads, while holding the write lock. VS Code then issues get connection requests for the CCloud connection, which blocks other event loop threads, and eventually, all event loop threads get blocked with the sidecar becomeing completely unresponsive (very sad). Once the 5 minute timeout (unsure where in the Java stack this is set) is hit, the originally blocked thread releases the lock, allowing _all_ other threads to resume.

Follow up to the DNS issue: We need to reduce client timeout to something like 10 seconds when issuing HTTP request to check authentication status. See #106

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/auth/CCloudOAuthContext.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/auth/CCloudOAuthContext.java
@@ -120,12 +120,6 @@ public class CCloudOAuthContext implements AuthContext {
         return Future.failedFuture(new CCloudAuthenticationFailedException(errorMessage));
       }
 
-      try {
-        Thread.sleep(5000);
-      } catch (InterruptedException e) {
-        Thread.currentThread().interrupt();
-      }
-
       return webClientFactory.getWebClient()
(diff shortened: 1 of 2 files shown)


MSeal (approved):
Looks to cover the issue in the way described. A little hard to know the race is resolved without a (difficult to write) test or running in through it's paces in a repeating conditional state.

rhauch (approved):
LGTM. Will we also have a PR that cherry-picks this to `main`?

rohitsanj:
> LGTM. Will we also have a PR that cherry-picks this to `main`?

Yes. I wanted to get this PR merged before raising the `main` cherry-pick PR.

PR #115, opened 2024-10-24
Retry get topic after creating topic

### LINKED ISSUE #114: Retry get topic after creating in create topic Kafka REST implementation
We try to get the topic as soon as it is created in our internal Kafka REST implementation - https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/kafkarest/TopicManagerImpl.java#L53-L57, this has been shown to sometimes fail in our tests and adding a retry here will make the implementation more robust. 

This is recommended in the Javadoc for `UnknownTopicOrPartitionException` as well, "This exception is retriable because the topic or partition might subsequently be created." (https://kafka.apache.org/20/javadoc/org/apache/kafka/common/errors/UnknownTopicOrPartitionException.html)

rohitsanj:
## Summary of Changes

Resolves #114 

This first manifested as flakiness in our integration tests. We need to retry the get topic operation after creating a topic.

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/kafkarest/TopicManagerImpl.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/kafkarest/TopicManagerImpl.java
@@ -64,7 +64,7 @@ public class TopicManagerImpl implements TopicManager {
             .retry()
             .atMost(3)
         );
-    }
+  }
 
   @Override
   public Uni<Void> deleteKafkaTopic(String clusterId, String topicName) {
--- a/src/main/java/io/confluent/idesidecar/restapi/kafkarest/TopicManagerImpl.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/kafkarest/TopicManagerImpl.java
@@ -15,6 +15,7 @@ import java.util.function.Supplier;
 import org.apache.kafka.clients.admin.DescribeTopicsOptions;
 import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.clients.admin.TopicDescription;
+import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/idesidecar/restapi/kafkarest/TopicManagerImpl.java
```diff
@@ -54,7 +55,15 @@ public Uni<TopicDescription> createKafkaTopic(
...
             createTopicRequestData.getTopicName(),
             false
-        ));
+            )
+            // The topic may not be immediately available after creation, so we retry a few times
+            // This is also recommended in the Javadoc for UnknownTopicOrPartitionException
+            // "This exception is retryable because the topic or partition might
+            // subsequently be created."
+            .onFailure(UnknownTopicOrPartitionException.class)
+            .retry()
```
rhauch, line 64: Can we add a delay for the retries?

rohitsanj: Added exp backoff

src/main/java/io/confluent/idesidecar/restapi/kafkarest/TopicManagerImpl.java
```diff
@@ -54,7 +56,18 @@ public Uni<TopicDescription> createKafkaTopic(
...
+            // Exponential backoff with a max of 3 retries
+            // Initial delay of 150ms, max delay of 1s
+            .withBackOff(Duration.ofMillis(150), Duration.ofMillis(1000))
```
rhauch, line 68: Let's merge this as-is (this is only a test), but in the future I'd do something like to make it more readable:
```suggestion
            .withBackOff(Duration.ofMillis(150), Duration.ofSeconds(1))
```

rhauch (approved):
LGTM. Thanks for adding the backoff.

PR #120, opened 2024-10-25
Convert from Checkstyle properties file to XML

### LINKED ISSUE #119: Convert from Checkstyle properties file to XML
Checkstyle has moved to [XML files for configuration](https://checkstyle.sourceforge.io/config.html). We could continue to use properties file, but XML is better documented.

rhauch:
Resolves #119

## Summary of Changes
Checkstyle has moved to XML files for configuration. We could continue to use properties file, but XML is better documented. See https://checkstyle.sourceforge.io/config.html

I've verified the new XML config file is picked up by temporarily adding a non-final class with a private default constructor, and checkstyle fails on that. Without that change, though, the build works as expected.


rohitsanj (approved):
LGTM, thanks!

PR #131, opened 2024-10-30
Fix the Connection API and Template API resource types to support deserializing

rhauch:
## Summary of Changes

The `ConnectionList` and `TemplateList` did not fully support deserializing, since production code doesn’t require it. However, this is needed when using REST Easy to deserialize these API responses with:
```
var connectionsList = given()
        .when()
        .get()
        .then()
        .statusCode(200)
        .contentType(ContentType.JSON)
        .extract().body()as(ConnectionsList.class);
```
Prior to this change, most of our tests extracted the body as a JSON representation, then parsed it into `JsonNode`, and manipulated the result. But extracting the body into its deserialized form is far simpler.

This PR does **not** otherwise change API behavior or the OpenAPI specification, and only changes how the Connection API and Template API resource types are implemented and tested. Specifically, it refactors the `BaseList`, `BaseModel` classes and their subtypes to be more conventional (e.g., storing the serialized fields, rather than deriving serialized fields transiently from other hidden fields) and adds new creator methods and additional annotations. The PR changes a few tests to use the type-specific deserialization approach, and adds new tests to verify the types could be deserialized, serialized, and then deserialized again without losing any information. These changes make it very easy to add in the future new tests that use the type-specific deserialization.

This PR does not rewrite the numerous tests that use the JSON approach.

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/featureflags/HttpFlagEvaluationProvider.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/featureflags/HttpFlagEvaluationProvider.java
@@ -14,6 +14,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.function.Consumer;
+import org.jboss.logging.Logger;
 
 /**
  * A {@link FeatureProject.Provider} that makes a remote HTTP call to LaunchDarkly APIs to
@@ -26,6 +27,7 @@ class HttpFlagEvaluationProvider implements FeatureProject.Provider {
   final ObjectMapper objectMapper;
   final String clientId;
   final String fetchUri;
+  final boolean testMode;
 
   HttpFlagEvaluationProvider(
       @NotNull String projectName,
@@ -37,6 +39,7 @@ class HttpFlagEvaluationProvider implements FeatureProject.Provider {
     this.clientId = clientId;
     this.fetchUri = fetchUri;
     this.objectMapper = objectMapper;
+    this.testMode = fetchUri.startsWith("http://localhost");
   }
 
   @Override
@@ -81,9 +84,11 @@ class HttpFlagEvaluationProvider implements FeatureProject.Provider {
           })
           .onFailure(failure -> {
             if (failure instanceof FeatureFlagFailureException) {
+              var level = testMode ? Logger.Level.DEBUG : Logger.Level.ERROR;
               // This occurs when we're unable to parse the evaluation response or error response
               // from the provider. We DO want to log these, as the problem needs to be fixed.
-              Log.errorf(
+              Log.logf(
+                  level,
                   "Error evaluating feature flags for project '%s': %s",
                   projectName,
                   failure.getMessage(),
--- a/src/main/java/io/confluent/idesidecar/restapi/featureflags/HttpFlagEvaluationProvider.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/featureflags/HttpFlagEvaluationProvider.java
@@ -8,6 +8,7 @@ import io.confluent.idesidecar.restapi.exceptions.FeatureFlagFailureException;
 import io.confluent.idesidecar.restapi.util.WebClientFactory;
 import io.quarkus.logging.Log;
 import io.quarkus.runtime.annotations.RegisterForReflection;
+import io.quarkus.runtime.configuration.ConfigUtils;
 import io.smallrye.common.constraint.NotNull;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
@@ -23,11 +24,12 @@ import org.jboss.logging.Logger;
 @RegisterForReflection
 class HttpFlagEvaluationProvider implements FeatureProject.Provider {
 
+  static final boolean TEST_MODE = ConfigUtils.getProfiles().contains("test");
+
   final String projectName;
   final ObjectMapper objectMapper;
   final String clientId;
   final String fetchUri;
-  final boolean testMode;
 
   HttpFlagEvaluationProvider(
       @NotNull String projectName,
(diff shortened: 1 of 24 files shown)


src/main/java/io/confluent/idesidecar/restapi/featureflags/HttpFlagEvaluationProvider.java
```diff
@@ -81,9 +84,11 @@ public void evaluateFlags(
           })
           .onFailure(failure -> {
             if (failure instanceof FeatureFlagFailureException) {
+              var level = testMode ? Logger.Level.DEBUG : Logger.Level.ERROR;
```
flippingbits, line 87: Would it make sense to turn the log level into an application configuration that we can set in the [application.yml](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/resources/application.yml) so that we can use different log levels for different modes/profiles? I feel we should not implement custom code for detecting the _current_ profile/mode.

src/main/java/io/confluent/idesidecar/restapi/resources/ConnectionsResource.java
```diff
@@ -60,10 +60,12 @@ public Uni<ConnectionsList> listConnections() {
                   .completionStage(() -> getConnectionModel(connection.getSpec().id())))
               .collect(Collectors.toList());
           if (connectionFutures.isEmpty()) {
+            Log.error("Returning no connections");
```
flippingbits, line 63: I think this does not warrant the error level. What about changing the log level to debug?

src/main/java/io/confluent/idesidecar/restapi/resources/ConnectionsResource.java
```diff
@@ -60,10 +60,12 @@ public Uni<ConnectionsList> listConnections() {
...
             return Uni
                 .createFrom()
-                .item(ConnectionsList.from(List.of()));
+                .item(new ConnectionsList());
           }
+          Log.errorf("Returning %d connections", connectionFutures.size());
```
flippingbits, line 68: Same as above; what about changing the log level to debug?

src/test/resources/connections/list-connections-response.json
```diff
@@ -33,8 +36,7 @@
       "metadata":
       {
         "self": "http://localhost:26637/gateway/v1/connections/2",
-        "resource_name": null,
-        "sign_in_uri": "https://login.confluent-dev.io/oauth/authorize?response_type=code&code_challenge_method=S256&code_challenge=f0883e0c8257e80f15d993a26e498da93ce2e1922a04cdc54bce6a0de0127281&state=2&client_id=cUmAgrkbAZSqSiy38JE7Ya3i7FwXmyUF&redirect_uri=http://127.0.0.1:26637/gateway/v1/callback-vscode-docs&scope=email+openid+offline_access"
+        "resource_name": null
```
flippingbits, line 39: Do we want to include the attribute despite being `null`?

flippingbits (commented):
Great work, @rhauch! Thanks for working on making testing more straightforward. I have a few minor questions and suggestions.

src/main/java/io/confluent/idesidecar/restapi/featureflags/HttpFlagEvaluationProvider.java
rhauch: I found these log messages very distracting in tests, but they are important to have as errors in production. I'm not sure how to accomplish this by configuring the log levels for the whole application but not otherwise affecting something else.

src/main/java/io/confluent/idesidecar/restapi/resources/ConnectionsResource.java
rhauch: Ooops. This was a debugging statement. I'll remove.

src/main/java/io/confluent/idesidecar/restapi/resources/ConnectionsResource.java
rhauch: Same as above. It's not needed. I'll double check other log message changes.

src/test/resources/connections/list-connections-response.json
rhauch: FYI: I've found that removing the above with my other changes didn't fail some tests, because of a regression I incorporated into my code. I'm working on a fix locally, but will probably get to it tomorrow. Thanks for pointing this out.

src/main/java/io/confluent/idesidecar/restapi/featureflags/HttpFlagEvaluationProvider.java
flippingbits: That makes sense. What about determining if we're running in test mode by checking if `io.quarkus.runtime.configuration.ConfigUtils.getProfiles()` contains `test` instead of verifying if `fetchUri` starts with `"http://localhost"`? I feel that's a bit more native to Quarkus.

rhauch: Thanks for the suggestion. I followed your suggestion and used `getProfiles()`.

src/test/resources/connections/list-connections-response.json
rhauch: I ended up keeping the original code. The test cases don't compare the `sign_in_uri` values anyway (since one component of the URL changes every time), and my new assertion logic ensures actual and expected values are either both null or both non-blank.

flippingbits (approved):
Thanks for the update. LGTM!

PR #134, opened 2024-10-31
Refactor the sidecar integration tests to support different types of connections

rhauch:
## Summary of Changes

Previously, we had a `ConfluentLocalTestBed` class that managed containers for Kafka and SR, and had utility functions (and state) for working with those containers and the sidecar process being tested.

This design had a few issues. First, it mixed together multiple things. Second, it required classes extend the `ConfluentLocalTestBed` class that would have made it difficult to use a different environment, say for integration testing. Third, it made it difficult to run a suite of tests on local connections AND on direct connections. Relying primarily upon inheritance is restrictive in Java due to the single base class approach.

This PR refactors the `ConfluentLocalTestBed` into two distinct components:
* a `LocalTestEnvironment` that extends a new`TestEnvironment` interface, used to manage containers for a Kafka cluster and Schema Registry cluster
* a `SidecarClient` that provides useful functions for working with a sidecar process that is being tested.

Instead of tests extending the `ConfluentLocalTestBed`, this PR introduces a new `AbstractSidecarIT` class that extends the `SidecarClient` to make the methods to operate on the sidecar more easily accessed by tests. More importantly related to local and direct connections, there is a set of connection-initialization methods that test classes can use in their `@BeforeEach` methods. This will make it far easier to extend existing IT classes to run the tests against direct connections.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


rhauch:
The build times on this are way up. I think more of the integration tests are running during unit testing w/ surefire rather than only during integration test phase w/ failsafe. I'm still looking into this.

rhauch:
The changes introduced in this PR significantly added to the build time -- it was taking 30 minutes to do a regular build. It turns out that I was not reusing containers as much as I had planned, and the extra work to set up and tear down the connections (and issuing multiple GraphQL queries for each) after every test was significantly time consuming.

Prior to this change, every `AbstractSidecarIT` would start its own containers and shut them down after the subclass' tests were run. This added about 5-10 seconds per subclass, and since we currently have 7 subclasses, this is over minute just to *restart* and shutdown the _extra_ containers. And will likely have more subclasses in the future.

Also, all of the tests in those IT test classes ended up creating a new local connection, which could add 3-5 seconds to each test. Part of this was due to the way in which I was using GraphQL queries during startup to find the clusters and during shutdown to find the clusters that were used. And soon we’ll double the tests by running those tests against direct connections. This was totally unacceptable.

This latest commit improves the `AbstractSidecarIT` base class to efficiently reuse containers and connections across many tests. First, the `AbstractSidecarIT` base class starts up the `LocalTestEnvironment` containers before the first integration test is run, and then shut them down after *all* integration tests have completed. This is true no matter how many subclasses of `AbstractSidecarIT` there are. Oh, and all of the subjects and topics created in one tests are cleaned up between each test.

Second, this PR introduces the concept of a “test scope” to the creation of connections, so that connections can be created once and used for all of the integration tests that use the same scope. All of the `AbstractSidecarIT` subclasses use a test scope named for the concrete IT class, though the technique is flexible to allow other patterns. This change can save up to five minutes across a hundred tests.

Note that this allows us to add other `TestEnvironment` instances (e.g., for CP or CCloud) to `AbstractSidecarIT` in the future, if we want to create them. Any containers used in those `TestEnvironment` instances would also be _started once and reused for all integration tests_.

src/main/java/io/confluent/idesidecar/restapi/filters/AccessTokenFilter.java
```diff
@@ -81,7 +81,11 @@ public void enforceAccessToken(RoutingContext rc) {
 
         String expectedToken = accessTokenBean.getToken();
         if (expectedToken == null) {
-          Log.error("Request for " + requestPath + " made before handshake route hit! Rejecting.");
+          if (enabled.get()) {
+            Log.error("Request for " + requestPath + " made before handshake route hit! Rejecting.");
+          } else {
+            Log.debug("Request for " + requestPath + " made without handshake.");
+          }
```
rhauch, line 88: This is another log message that shows up for every API call made in our integration tests (where `@TestProfile(NoAccessFilterProfile.class)` is used), and inaccurately says the request is rejected. Again, the production log messages are very useful, but are not nearly as important during integration tests.

src/test/java/io/confluent/idesidecar/restapi/util/AbstractSidecarIT.java
```diff
@@ -0,0 +1,241 @@
...
+ * connections for a specified "test scope". This allows tests to use _different_ connections,
+ * but reusing connections depends on which tests reuse the same scopes.
+ *
+ * <p>And yet other integration tests may want each test to set up the connections it needs.
+ * In this case, each test should call the appropriate {@code setup*Connection} method and, when
+ * finished, call {@link #deleteAllConnections(Predicate)} to clean up the test's connections.
+ * This approach is the most flexible, but also the slowest, as each test will need to create
+ * the connections it needs.
+ */
+public abstract class AbstractSidecarIT extends SidecarClient {
```
rhauch, line 57: The `ConfluentLocalTestBed` class was broken up into this class, the `SidecarClient` class that provides easy-to-use functions that tests can use to submit requests to the sidecar APIs, and the `LocalTestEnvironment` for managing the connections.

src/test/java/io/confluent/idesidecar/restapi/kafkarest/api/ClusterV3ApiImplIT.java
```diff
@@ -3,18 +3,24 @@
...
 @Tag("io.confluent.common.utils.IntegrationTest")
 @TestProfile(NoAccessFilterProfile.class)
-public class ClusterV3ApiImplIT extends ConfluentLocalTestBed {
+public class ClusterV3ApiImplIT extends AbstractSidecarIT {
+
+  @BeforeEach
+  public void beforeEach() {
+    setupLocalConnection(ClusterV3ApiImplIT.class);
+  }
+
```
rhauch, line 23: This is all that changes within (most) of the integration test classes.

rhauch (commented):
A few notes on the changes in this PR. Nearly all of the changes are in the test codebase, and should not affect the production behavior of the sidecar.

flippingbits (dismissed):
Excellent work, @rhauch! 💯 I have only a few very minor suggestions/comments. Feel free to decide to address them in this PR or a follow-up PR, if at all.

src/test/java/io/confluent/idesidecar/restapi/util/SidecarClient.java
```diff
@@ -0,0 +1,680 @@
...
+        } else {
+          assertEquals(200, createSchemaVersionResp.statusCode());
+        }
+
+        // Sleep for a bit to allow the schema to be registered
+        try {
+          Thread.sleep(50);
+        } catch (InterruptedException e) {
+          // Ignore
+        }
```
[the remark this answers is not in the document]
rhauch: That's a good suggestion. I logged #139 and marked it as `good first issue` and P2.

src/test/java/io/confluent/idesidecar/restapi/util/LocalTestEnvironment.java
```diff
@@ -0,0 +1,108 @@
...
+import java.time.Duration;
+
+/**
+ * A {@link TestEnvironment} that starts a local Confluent Local container with Kafka broker and
+ * Kafka REST server, and a Confluent Platform Schema Registry container.
+ */
+@TestProfile(NoAccessFilterProfile.class)
+public class LocalTestEnvironment implements TestEnvironment {
+
+  private static final String KAFKA_INTERNAL_LISTENER = "PLAINTEXT://confluent-local-broker-1:29092";
```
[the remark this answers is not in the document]
rhauch: Good idea. I logged #140

rhauch:
Thanks for the review, @flippingbits. I've incorporated most of your changes, and created a few issues for the others.

flippingbits (approved):
Thanks a lot for creating the issues and updating the PR! LGTM.

PR #135, opened 2024-10-31
Fix the native image not including the `sign_in_uri` on CCloud connections

rhauch:
Not sure why `metadata` in the response of a newly-created CCloud connections was empty, but only in the native image (worked fine in dev-mode and in the unit+integration tests).

To fix, I added `@RegisterForReflection` to all of the resource classes that extend any of the `BaseList` or `BaseModel`. Probably over-cautious, but this fixes the problem.

I’ll add an integration test in a subsequent PR to try and catch this kind of regression in the future.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


rohitsanj (approved):
Thanks, LGTM!

PR #148, opened 2024-11-07
Consider baked-in default certs when using PEM file

flippingbits:
When using a PEM file to verify the server-side TLS certs, make sure that we do not lose access to the default certs baked into the native executable.

We do not expect this change to introduce a regression because we won't skip any certs but add a few additional ones.

Fixes https://github.com/confluentinc/vscode/issues/523

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


rhauch (approved):
LGTM

Issue #142, opened 2024-11-05
Clear cluster cache on connection update events


rohitsanj:
Currently, when the SR URI of a local connection is updated to a new value, the sidecar continues to use the old value due to the cluster cache having stale information. 

The [`onConnectionUpdated` event handler method](https://github.com/confluentinc/ide-sidecar/blob/0d355a0334e0d525968faf2960a512a9b73f9da1/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java#L192-L194) needs to include logic to clear all cached clusters for this connection, similar to what we do in the [`onConnectionDeleted` event listener method](https://github.com/confluentinc/ide-sidecar/blob/0d355a0334e0d525968faf2960a512a9b73f9da1/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java#L211-L212). 

Simply clearing the cluster cache would do the trick by forcing future Kafka/SR REST proxy requests to fetch and re-cache any new cluster details.

This will need an associated test as well to prevent regressions.

Cerchie:
addressed here https://github.com/confluentinc/ide-sidecar/pull/150

PR #150, opened 2024-11-07
Fix/clear cluster cache on connection update

Cerchie:
## Summary of Changes

Addresses #142 by clearing cache on connection update and adding test

## Pull request checklist

- Tests:
    - [x] Added new
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
@@ -2,21 +2,15 @@ package io.confluent.idesidecar.restapi.cache;
 
 import graphql.VisibleForTesting;
 import io.confluent.idesidecar.restapi.connections.ConnectionState;
+import io.confluent.idesidecar.restapi.connections.ConnectionStates;
 import io.confluent.idesidecar.restapi.events.ClusterKind;
 import io.confluent.idesidecar.restapi.events.Lifecycle;
 import io.confluent.idesidecar.restapi.exceptions.ClusterNotFoundException;
 import io.confluent.idesidecar.restapi.exceptions.ConnectionNotFoundException;
 import io.confluent.idesidecar.restapi.models.ClusterType;
+import io.confluent.idesidecar.restapi.models.ConnectionSpec;
 import io.confluent.idesidecar.restapi.models.ConnectionSpec.ConnectionType;
-import io.confluent.idesidecar.restapi.models.graph.CCloudKafkaCluster;
-import io.confluent.idesidecar.restapi.models.graph.CCloudSchemaRegistry;
-import io.confluent.idesidecar.restapi.models.graph.Cluster;
-import io.confluent.idesidecar.restapi.models.graph.ClusterEvent;
-import io.confluent.idesidecar.restapi.models.graph.KafkaCluster;
-import io.confluent.idesidecar.restapi.models.graph.RealCCloudFetcher;
-import io.confluent.idesidecar.restapi.models.graph.RealDirectFetcher;
-import io.confluent.idesidecar.restapi.models.graph.RealLocalFetcher;
-import io.confluent.idesidecar.restapi.models.graph.SchemaRegistry;
+import io.confluent.idesidecar.restapi.models.graph.*;
 import io.confluent.idesidecar.restapi.util.TimeUtil;
 import io.quarkus.logging.Log;
 import io.smallrye.common.constraint.NotNull;
@@ -192,8 +186,11 @@ public class ClusterCache {
   void onConnectionUpdated(@ObservesAsync @Lifecycle.Updated ConnectionState connection) {
     Log.infof("Updated %s", connection.getSpec());
 
-    // clear the cache to force future Kafka/SR REST proxy requests to fetch and re-cache any new cluster details
-    clustersByConnectionId.remove(connection.getSpec().id());
+    // Replace the existing cache for this connection with a new one
+    clustersByConnectionId.put(
+            connection.getId(),
+            new Clusters(connection.getId(), connection.getType())
+    );
   }
 
   void onConnectionEstablished(
--- a/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
@@ -1,6 +1,5 @@
 package io.confluent.idesidecar.restapi.cache;
 
-
 import graphql.VisibleForTesting;
 import io.confluent.idesidecar.restapi.connections.ConnectionState;
 import io.confluent.idesidecar.restapi.events.ClusterKind;
--- a/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
@@ -193,7 +193,7 @@ public class ClusterCache {
     Log.infof("Updated %s", connection.getSpec());
 
     // clear the cache to force future Kafka/SR REST proxy requests to fetch and re-cache any new cluster details
-    clustersByConnectionId.clear();
+    clustersByConnectionId.remove(connection.getSpec().id());
   }
 
   void onConnectionEstablished(
--- a/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
@@ -191,6 +191,9 @@ public class ClusterCache {
 
   void onConnectionUpdated(@ObservesAsync @Lifecycle.Updated ConnectionState connection) {
     Log.infof("Updated %s", connection.getSpec());
+
+    // clear the cache to force future Kafka/SR REST proxy requests to fetch and re-cache any new cluster details
+    clustersByConnectionId.clear();
   }
 
   void onConnectionEstablished(
--- a/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
@@ -1,16 +1,23 @@
 package io.confluent.idesidecar.restapi.cache;
 
+
 import graphql.VisibleForTesting;
 import io.confluent.idesidecar.restapi.connections.ConnectionState;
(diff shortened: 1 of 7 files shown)


src/test/java/io/confluent/idesidecar/restapi/cache/ClusterCacheTest.java
```diff
@@ -224,6 +225,16 @@ void shouldFindAlreadyLoadedSchemaRegistryForConnectionAndLazilyLoadKafkaCluster
...
+    @Test
+    void shouldClearCacheOnConnectionUpdateAndReCacheDetails() {
+      // When a connection is updated
+      cache.onConnectionUpdated(CONNECTION_1);
+
+      // assert cache is empty
+      assertEquals(0, cache.clustersByConnectionId.size());
+
+    }
+
```
rhauch, line 237: I think we essentially want to do something like this:
```suggestion
    @Test
    void shouldRemoveCacheOnConnectionUpdate() {
      // When two connections are created
      cache.onConnectionCreated(CONNECTION_1);
      cache.onConnectionCreated(CONNECTION_2);

      // and a Kafka cluster is loaded in each connection
      cache.onLoadingKafkaCluster(
          new ClusterEvent(CONNECTION_1_ID, ConnectionSpec.ConnectionType.CCLOUD, LKC_1)
      );
      cache.onLoadingKafkaCluster(
          new ClusterEvent(CONNECTION_2_ID, ConnectionSpec.ConnectionType.CCLOUD, LKC_2)
      );

      // Then there will be a connection cache for both connections
      assertConnection(CONNECTION_1);
      assertConnection(CONNECTION_2);

      // and the cache for each connection contains their Kafka cluster
      assertKafkaCluster(LKC_1);
      assertKafkaCluster(LKC_2);

      // When one of the connections is updated
      cache.onConnectionUpdated(CONNECTION_1);

      // then the cache for the connection will have been cleared (no cluster will be found)
      var clusters = assertConnection(CONNECTION_1);
      assertThrows(
          ClusterNotFoundException.class,
          () -> clusters.getKafkaCluster(LKC_1.id(), false)
      );

      // but the cache for the other connection is unaffected
      assertConnection(CONNECTION_2);
      assertKafkaCluster(LKC_2);

      // and the Kafka cluster for connection 1 will be loaded again
      expectFindKafkaCluster(CONNECTION_1_ID, LKC_1.id(), LKC_1);
      // when the Kafka cluster is accessed again
      assertEquals(LKC_1, clusters.getKafkaCluster(LKC_1.id()));
    }
```
This does a couple of things. First, it verifies that the cache for a connection is cleared when it is updated. We're not quite checking that it is cleared, but we're verifying that a cluster that used to be there no longer is in the cache. We could add an `isEmpty()` protected method to the `Clusters` class, and also add an assertion in the test that it is empty.

Second, it verifies that other connections are not affected. The cluster that was loaded into the cache for connection 2 is still there after connection 1 is updated.

Third, it verifies that after a connection is updated and it's cache is cleared (or technically recreated), then the cache for connection 1 will reload the Kafka cluster if it is queried.

BTW, here are the additional constants that are needed for `CONNECTION_2` and `LKC_2`:
```
    static final String CONNECTION_2_ID = "c2";
    static final ConnectionState CONNECTION_2 = ConnectionStates.from(
        new ConnectionSpec(
            CONNECTION_2_ID,
            "Conn2",
            ConnectionSpec.ConnectionType.CCLOUD
        ),
        null
    );
    static final CCloudOrganization ORG_2 = new CCloudOrganization(
        "org2",
        "my-other-org",
        false,
        CONNECTION_2_ID
    );
    static final CCloudEnvironment ENV_2 = new CCloudEnvironment(
        "env-456",
        "my-env-2",
        CCloudGovernancePackage.ESSENTIALS
    ).withOrganization(ORG_2).withConnectionId(CONNECTION_2_ID);
    static final CCloudKafkaCluster LKC_2 = new CCloudKafkaCluster(
        "lkc-2",
        "my-other-kafka-cluster",
        CloudProvider.AWS,
        "us-west-2",
        "pkc-456"
    ).withEnvironment(ENV_2).withOrganization(ORG_2).withConnectionId(CONNECTION_2_ID);
```

src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
```diff
@@ -191,6 +191,9 @@ void onConnectionCreated(@ObservesAsync @Lifecycle.Created ConnectionState conne
 
   void onConnectionUpdated(@ObservesAsync @Lifecycle.Updated ConnectionState connection) {
     Log.infof("Updated %s", connection.getSpec());
+
+    // clear the cache to force future Kafka/SR REST proxy requests to fetch and re-cache any new cluster details
+    clustersByConnectionId.remove(connection.getSpec().id());
```
rhauch, line 196: If we just remove it, then when the cache is used to find clusters for a particular connection, that connection will not be found and we'll get an exception.

I think we actually want to _replace_ the connection's cache:
```suggestion

    // Replace the existing cache for this connection with a new one
    clustersByConnectionId.put(
        connection.getId(),
        new Clusters(connection.getId(), connection.getType())
    );
```
BTW, we could also have changed `Clusters` to have a `clear()` method, but that's not as good as _replacing_ the `Clusters` object for the connection. Why? The `Clusters` class has two maps, and it's not possible to atomically clear both. (Another thread might see an inconsistent `Clusters` object between our clearing of the two maps.) However, `clustersByConnectionId` is a concurrent map, and replacing the existing value on that entry requires changing one reference, which is atomic.

rhauch (commented):
Thanks, @Cerchie . Based on your questions offline, I've given some suggestions on how to better implement the clearing and test to verify the desired side effects take effect and no undesired side effects take effect. Let me know if this makes sense.

rhauch (commented):
Thanks for the changes, @Cerchie. I have a few more suggestions, plus one question for @flippingbits and @rohitsanj below.

src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java
```diff
@@ -2,21 +2,15 @@
...
-import io.confluent.idesidecar.restapi.models.graph.CCloudKafkaCluster;
-import io.confluent.idesidecar.restapi.models.graph.CCloudSchemaRegistry;
-import io.confluent.idesidecar.restapi.models.graph.Cluster;
-import io.confluent.idesidecar.restapi.models.graph.ClusterEvent;
-import io.confluent.idesidecar.restapi.models.graph.KafkaCluster;
-import io.confluent.idesidecar.restapi.models.graph.RealCCloudFetcher;
-import io.confluent.idesidecar.restapi.models.graph.RealDirectFetcher;
-import io.confluent.idesidecar.restapi.models.graph.RealLocalFetcher;
-import io.confluent.idesidecar.restapi.models.graph.SchemaRegistry;
+import io.confluent.idesidecar.restapi.models.graph.*;
```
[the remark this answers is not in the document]
Cerchie: right you had mentioned this in an earlier PR. actually not sure how this one snuck in here, but I'm removing 🧹

src/main/java/io/confluent/idesidecar/restapi/messageviewer/MessageViewerClusterInfoProcessor.java
```diff
@@ -90,6 +90,15 @@ void onConnectionDisconnected(
...
 
+  /**
+   * Respond to the connection being updated by clearing cached information.
+   *
+   * @param connection the connection that was prior to update
+   */
+  void onConnectionUpdated(@ObservesAsync @Lifecycle.Updated ConnectionState connection) {
+    clearCaches();
+  }
+
```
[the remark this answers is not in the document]
rohitsanj: Thanks for catching that. I've created https://github.com/confluentinc/ide-sidecar/issues/153.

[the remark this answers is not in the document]
rhauch: @Cerchie let's remove the change to this class for now, per the above.

rhauch (commented):
Thanks, @Cerchie. Just a few more minor suggestions.

src/test/java/io/confluent/idesidecar/restapi/cache/ClusterCacheTest.java
```diff
@@ -32,6 +14,10 @@
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.*;
```
rhauch, line 17: Please remove the wildcard import.

src/test/java/io/confluent/idesidecar/restapi/cache/ClusterCacheTest.java
```diff
@@ -1,28 +1,11 @@
...
-import io.confluent.idesidecar.restapi.models.graph.CCloudGovernancePackage;
-import io.confluent.idesidecar.restapi.models.graph.CCloudKafkaCluster;
-import io.confluent.idesidecar.restapi.models.graph.CCloudOrganization;
-import io.confluent.idesidecar.restapi.models.graph.CCloudSchemaRegistry;
-import io.confluent.idesidecar.restapi.models.graph.CloudProvider;
-import io.confluent.idesidecar.restapi.models.graph.ClusterEvent;
-import io.confluent.idesidecar.restapi.models.graph.KafkaCluster;
-import io.confluent.idesidecar.restapi.models.graph.RealCCloudFetcher;
-import io.confluent.idesidecar.restapi.models.graph.RealLocalFetcher;
+import io.confluent.idesidecar.restapi.models.graph.*;
```
rhauch, line 8: Please remove the wildcard import.

rhauch (commented):
Thanks, @Cerchie. Two more wildcard imports need to be removed, but otherwise looks great.

src/test/java/io/confluent/idesidecar/restapi/cache/ClusterCacheTest.java
Cerchie: removing! also did [this ](https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports)so these shouldn't appear anymore

rhauch (approved):
Very nice work, @Cerchie!

Issue #123, opened 2024-10-26
Update connection `status` to work for local, CCloud and direct connections


rhauch:
Continuation of #96 and #122, and followup to #116 that added initial and limited direct connections (with GraphQL support) but did not support credentials nor unify the `status` to work for all kinds of connections. (The `status` is heavily oriented around CCloud connections.)

Note that direct connections do not yet support credentials, so direct connections will only work with Kafka clusters and SR that have no authentication. We'll track adding support for other credential types in subsequent issues:
* basic authN (#124)
* mTLS authN (#125)
* OAuth authN (#126)
* Kerberos authN, or SSL/GSSAPI (#127)

rhauch:
Every `Connection` resource includes a `spec` field and `status` field. The `spec` is the user-defined information about the connection, and the `status` is the system-generated information. The `status` field is defined in the OpenAPI spec as the `ConnectionStatus` schema:
```
    ConnectionStatus:
      required:
      - authentication
      type: object
      properties:
        authentication:
          $ref: "#/components/schemas/Authentication"
```
with the `Authentication` schema:
```
    Authentication:
      required:
      - status
      type: object
      properties:
        status:
          $ref: "#/components/schemas/Status"
        requires_authentication_at:
          $ref: "#/components/schemas/Instant"
        user:
          $ref: "#/components/schemas/UserInfo"
        errors:
          $ref: "#/components/schemas/AuthErrors"
```
and the `Status` schema:
```
    Status:
      enum:
      - NO_TOKEN
      - VALID_TOKEN
      - INVALID_TOKEN
      - FAILED
      type: string
```

New connection resources always have an "initial" status of (in a simplified not-quite-JSON representation):
```
status: {
  authentication: {
    status: NO_TOKEN
  }
}
```
and successfully-connected CCloud connections have this status:
```
status: {
  authentication: {
    status: SUCCESS,
    user_info: ...
    errors: ...
  }
}
```
The `errors` field is used only if the status is FAILED or INVALID_TOKEN.

This design was added for CCloud connections and the structure and terminology of the `Status` enum literals is very much oriented around CCloud and/or OAuth. 

Meanwhile, the `spec` for connection resources is defined in the OpenAPI spec as:
```
    ConnectionSpec:
      description: The connection details that can be set or changed.
      type: object
      properties:
        id:
          description: The unique identifier of the connection resource.
          maxLength: 64
          minLength: 1
          type: string
        name:
[...]

PR #162, opened 2024-11-14
Evolve the schema of the Connection resource `status`

rhauch:
Part 1 of 2 to address #123 

## Summary of Changes

Change the `status` object in `Connection` resources to able to represent the statuses of the different external clusters/resources that connections can define.

The `ConnectionStatus` OpenAPI spec schema now looks like:
```
    ConnectionStatus:
      required:
      - authentication
      type: object
      properties:
        ccloud:
          $ref: "#/components/schemas/CCloudStatus"
        kafka_cluster:
          $ref: "#/components/schemas/KafkaClusterStatus"
        schema_registry:
          $ref: "#/components/schemas/SchemaRegistryStatus"
        authentication:
          $ref: "#/components/schemas/Authentication"
```
The `status.authentication` object is unchanged, but is being deprecated and replaced with the `status.ccloud` object. The `status.ccloud`, `status.kafka_clsuter` and `status.schema_registry` nested objects all use the `ConnectedState` enum:
```
    ConnectedState:
      enum:
      - NONE
      - ATTEMPTING
      - SUCCESS
      - EXPIRED
      - FAILED
      type: string
```
and there is a direct mapping from the older `Status` enumeration to this new enum.

The `status.authentication` object is being generated from the `status.ccloud` object, and the sidecar code has already transitioned to only using `status.ccloud`; the `status.kafka_cluster` and `status.schema_registry` objects will be set in a subsequent PR.

We will keep the `status.authentication` object for a time being, so the extension has time to migrate to the `status.ccloud`, `status.kafka_cluster` and `status.schema_registry` objects. Once that migration has been performed, we will remote the `status.authentication` object altogether, in a future PR.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/models/ConnectionStatus.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/models/ConnectionStatus.java
@@ -6,21 +6,145 @@ import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import io.confluent.idesidecar.restapi.auth.AuthErrors;
 import io.confluent.idesidecar.restapi.models.ConnectionStatus.Authentication.Status;
+import io.soabase.recordbuilder.core.RecordBuilder;
+import jakarta.validation.constraints.Null;
 import java.time.Instant;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
 /**
  * Represents the status of a connection.
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
-public record ConnectionStatus(@JsonProperty(required = true) Authentication authentication) {
+@JsonInclude(Include.NON_NULL)
+@RecordBuilder
+public record ConnectionStatus(
+    @JsonProperty("ccloud")
+    CCloudStatus ccloud,
+
+    @JsonProperty("kafka_cluster")
+    KafkaClusterStatus kafkaCluster,
+
+    @JsonProperty("schema_registry")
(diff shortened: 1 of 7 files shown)


src/main/java/io/confluent/idesidecar/restapi/models/ConnectionStatus.java
```diff
@@ -6,21 +6,145 @@
...
+
+    @JsonProperty("schema_registry")
+    SchemaRegistryStatus schemaRegistry
+) implements ConnectionStatusBuilder.With {
+
+  // TODO: Remove this once the extension has been updated to use the new status
+  @JsonProperty(required = true)
+  public Authentication authentication() {
+    if (ccloud != null) {
+    return new Authentication(
```
rohitsanj, line 35: Needs indentation fix

rohitsanj (commented):
Looks great! Just needs a small indentation fix.

rohitsanj:
Huh, I'd have expected CI to catch that indentation issue. I guess Checkstyle doesn't care about it.

src/main/java/io/confluent/idesidecar/restapi/models/ConnectionStatus.java
rhauch: Good catch. I'm surprised checkstyle didn't get this. We probably need to look into that; logged as #163

PR #165, opened 2024-11-15
Add `dry_run` flag to Produce Records API

### LINKED ISSUE #159: Add message validation functionality to Sidecar
Extend the produce records API to take a `?dryRun=true` query parameter and perform the following validations:

1. Does the schema version exist?
1. Does the specified subject exist? (Depends on #111)
1. Does the key content match the schema (if one is specified)?
1. Does the value content match the schema (if one is specified)?
1. Does the partition exist?

rohitsanj:
## Summary of Changes

Resolves #159 

The API change made in the PR to include a query parameter `dry_run` will not appear in the generated OpenAPI spec. The Internal Kafka REST routes are excluded from the generated OpenAPI spec by design. The query parameter will need to be documented in the [`ce-kafka-rest.openapi.yaml`](https://github.com/confluentinc/vscode/blob/main/src/clients/sidecar-openapi-specs/ce-kafka-rest.openapi.yaml) of confluentinc/vscode. (https://github.com/confluentinc/vscode/issues/613 tracks that work).

## Pull request checklist

- Tests:
    - [x] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/main/java/io/confluent/idesidecar/restapi/kafkarest/RecordsV3ApiImpl.java
```diff
@@ -63,18 +65,33 @@ public Uni<ProduceResponse> produceRecord(
...
+          // This should indicate that the request as a whole is valid.
+          .map(c -> ProduceResponse
+              .builder()
+              .clusterId(c.clusterId)
+              .topicName(c.topicName)
+              .build()
+          );
+    } else {
+      return uptoDryRun
+          .chain(this::sendSerializedRecord)
```
flippingbits, line 92: Do we need to chain this with `.onFailure(RuntimeException.class).transform(this::unwrapRootCause)`?

flippingbits (commented):
LGTM! WDYT of implementing the `dryRun` flag as an enum with the initial values `NONE` and `FULL`? For other endpoints, we might want to offer (partial) dry runs.

rohitsanj:
> WDYT of implementing the dryRun flag as an enum with the initial values NONE and FULL? For other endpoints, we might want to offer (partial) dry runs.

That makes sense! I'd like to keep it simple for now though since we don't need partial dry run for this particular endpoint.

src/main/java/io/confluent/idesidecar/restapi/kafkarest/RecordsV3ApiImpl.java
rohitsanj: Ah, yes. Thanks

flippingbits:
> I'd like to keep it simple for now though since we don't need partial dry run for this particular endpoint.

👍 Note that changing the query param from boolean to enum is a breaking change. OTOH, the VS Code extension is the only consumer of this API so we should be able to apply the change without causing harm if we communicate it clearly.

flippingbits (approved):

rohitsanj:
> Note that changing the query param from boolean to enum is a breaking change. OTOH, the VS Code extension is the only consumer of this API so we should be able to apply the change without causing harm if we communicate it clearly.

Thanks! I appreciate the well-reasoned approach.

src/main/java/io/confluent/idesidecar/restapi/kafkarest/RecordsV3ApiImpl.java
```diff
@@ -63,18 +65,35 @@ public Uni<ProduceResponse> produceRecord(
       @HeaderParam(CONNECTION_ID_HEADER) String connectionId,
       @PathParam("cluster_id") String clusterId,
       @PathParam("topic_name") String topicName,
+      @QueryParam("dry_run") @DefaultValue("false") boolean dryRun,
```
rhauch, line 68: If we're adding a query parameter, wouldn't the OpenAPI spec change?

rhauch (dismissed):
LGTM

flippingbits (approved):
LGTM. Thanks, @rohitsanj!

PR #166, opened 2024-11-15
Support producing records to local/direct Kafka connections without SR defined

rohitsanj:
## Summary of Changes

- Added tests to prove that the Produce Records endpoint works with local connections without SR defined. 
- Also added a test to prove that: Producing to a topic with schema details for a Kafka connection without SR fails with 400 bad request along with an appropriate error message prompting the user towards possible remedy: "This connection does not have an associated Schema Registry. Either define a Schema Registry for this connection or try again without specifying schema details: schema version, subject or subject name strategy."

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
```diff
@@ -28,13 +29,14 @@ public AdminClients() {
    * @param clusterId    The cluster ID
    * @return The AdminClient
    */
-  public AdminClient getClient(String connectionId, String clusterId) {
+  public AdminClient getClient(String connectionId, String clusterId)
+      throws ClusterNotFoundException {
+    // Generate the Kafka admin client configuration
+    final var config = configurator.getAdminClientConfig(connectionId, clusterId, false);
```
rohitsanj, line 35: Moved this out of the lambda function since it now raises a checked exception `ClusterNotFoundException`.

rhauch:
> Changed ClusterNotFoundException to be a checked exception instead of unchecked. This allows us to be more intentional about this exception propagating up the call stack, allowing us to possibly catch and ignore it at some level if appropriate.
> > Currently with it being an unchecked exception, it could be thrown in places where we actually expect not being able to find a cluster -- leading to bugs.

Actually, there are places where it should not be thrown, but because the exception would be a checked exception we'd have to catch the exception and decide how to handle -- possibly by wrapping it in a checked exception.

TBH, I'm not very enthusiastic about this particular change.

rhauch:
> ... since `KafkaProducer` themselves have no notion of Schema Registry configs. (A `KafkaProducer` may be passed a custom key/value serializer that looks up schemas in a Schema Registry and knows how to serialize a key/value into bytes, as described in this [blog](https://www.confluent.io/blog/how-schema-registry-clients-work/)).

Technically this is correct that the Java `KafkaProducer` has no notion of Schema Registry configs. However, `KafkaProducer` **does** have a notion of configuring key and value serializers by specifying class names and arbitrary properties that will be passed to the `Serializer` instances after instantiation (via [this line in KafkaProducer](https://github.com/apache/kafka/blob/b5158aa3ad90a844a7488addd2987a6acec61585/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L392)). And Schema Registry's implementations for the Avro, Protobuf and JSON Schema serializers **do** accept configuration properties with Schema Registry information (all SR-specific properties in the producer configuration should have the `schema.registry.` prefix). 

FWIW, the code linked above works as follows: the `KafkaProducer` instantiates the specified deserializer class with a no-arg constructor using the `config.getConfiguredInstance(...)` method, which ultimately calls [this protected method on AbstractConfig](https://github.com/apache/kafka/blob/b5158aa3ad90a844a7488addd2987a6acec61585/clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java#L384-L409) that, in the case of the Avro serializer, instantiates the `KafkaAvroSerializer` class and calls the `configure(...)` method with the producer's configuration properties. The `KafkaAvroSerializer` constructor [creates the `KafkaAvroSerializerConfig` object](https://github.com/confluentinc/schema-registry/blob/06cff6212c399ab7ec405f20b57f695883b0880c/avro-serializer/src/main/java/io/confluent/kafka/serializers/KafkaAvroSerializer.java#L49-L53) using the supplied properties, and this is responsible for configuring the serializer instance. The `KafkaAvroSerializerConfig` extends (ultimately) the [AbstractKafkaAvroSerDeConfig](https://github.com/confluentinc/schema-registry/blob/master/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroSerDeConfig.java) class that, as you can see, looks for all of the Schema Registry configuration properties that all begin with the `schema.registry.` prefix. This is also documented more generically and more ambiguously [here](https://docs.confluent.io/platform/current/schema-registry/schema_registry_onprem_tutorial.html#configuring-avro).

The blog you cited doesn't show this because the non-Java producer implementations don't work this way with the serializers, and typically require the client application to instantiate the serializers and pass them to the producer instance.

Of course, we're using the Java Kafka clients, so we want to keep the `includeSchemaRegistry` parameter in the `ClientConfigurator.getProducerClientConfig` method.

rhauch (approved):
Thanks for reverting the changes earlier. Looks good to me.

rhauch:
@rohitsanj, can you please update the PR description, in particular the first bullet that no longer reflects the changes in the PR:
> - Removed the `includeSchemaRegistry` parameter from `ClientConfigurator.getProducerClientConfig` since `KafkaProducer` themselves have no notion of Schema Registry configs. (A `KafkaProducer` may be passed a custom key/value serializer that looks up schemas in a Schema Registry and knows how to serialize a key/value into bytes, as described in this [blog](https://www.confluent.io/blog/how-schema-registry-clients-work/)).

src/test/java/io/confluent/idesidecar/restapi/kafkarest/RecordsV3Suite.java
```diff
@@ -199,17 +199,21 @@ static ArgumentSets validKeysAndValues() {
   @CartesianTest
   @CartesianTest.MethodFactory("validKeysAndValues")
   default void testProduceAndConsumeData(RecordData key, RecordData value) {
-    var topicName = randomTopicName();
+    produceAndConsume(this, key, value);
+  }
+
+  static <T extends ITSuite> void produceAndConsume(T test, RecordData key, RecordData value) {
```
rohitsanj, line 205: This is used in the `RecordsV3WithoutSRSuite`.

rohitsanj:
@rhauch I've adapted the *WithoutSR tests to the new integration test suite pattern. Please have a look, thanks!

.semaphore/semaphore.yml
```diff
@@ -162,7 +162,7 @@ blocks:
     task:
       agent:
         machine:
-          type: s1-prod-ubuntu20-04-arm64-1
+          type: s1-prod-ubuntu20-04-arm64-2
```
rhauch: Are you sure this is based off of the latest `main`? [This line in main](https://github.com/confluentinc/ide-sidecar/blob/08595b7f594d11eca31be9e95a3331dcfe38aad5/.semaphore/semaphore.yml#L165) is already changed to use `-2`.

src/test/java/io/confluent/idesidecar/restapi/kafkarest/RecordsV3Suite.java
rhauch: This is probably okay, but don't you think we probably want a common interface that defines the common (non-static) methods that could be used in `RecordsV3Suite` and `RecordsV3WithoutSRSuite`?

rhauch (commented):
Mostly looks good, but isn't it weird to add a static method that accepts the `ITSuite` subtype as a parameter? None of the other suites do this, so it seems like a bit of restructuring might help.

rohitsanj:
@rhauch Yeah, it was weird. I've refactored the common methods/static variables into an interface called `RecordsV3BaseSuite`, and made the other RecordsV3 interfaces extend this one.

rhauch (approved):
Thanks for refactoring the common methods into a base interface for the record test suites. LGTM!

PR #167, opened 2024-11-15
Make Consumer and Producer client config configurable

flippingbits:
## Summary of Changes

This change makes the configuration of the Kafka Consumer and Producer clients configurable. Also, similar to the Kafka Admin clients, it sets the `client.id` to a sidecar-specific value so that users can distinguish between consume/produce requests of the ide-sidecar and those of other Kafka applications.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

I manually verified the changes by setting up a Direct connection for a CCloud cluster and checking if the sidecar shows up on the Clients page:

Screenshot 2024-11-15 at 10 39 17


src/main/resources/application.yml
```diff
@@ -138,9 +138,19 @@ ide-sidecar:
...
   # (We may choose to allow connection-specific overrides in the future)
   admin-client-configs:
+    "client.id": Confluent for VS Code sidecar ${quarkus.application.version} - Admin
     "default.api.timeout.ms": 5000
     "request.timeout.ms": 5000
-    "client.id": Confluent for VS Code sidecar ${quarkus.application.version}
+  # Configuration set by the sidecar when instantiating the Kafka Consumer
+  # (We may choose to allow connection-specific overrides in the future)
+  consumer-client-configs:
+    "client.id": Confluent for VS Code sidecar ${quarkus.application.version} - Consumer
```
rhauch, line 147: I know this is what we've used for the admin client, but we need to be careful about using "VS Code" here, since the sidecar might be used in other IDEs. Also, do we want to include the sidecar version here, or the IDE's version? Do we need some unique portion of the client? What about including the device ID?

The SidecarInfo bean has this information, but we'd want to insert the values within the code, not here.

flippingbits: > I know this is what we've used for the admin client, but we need to be careful about using "VS Code" here, since the sidecar might be used in other IDEs.

Good point. We should use IDE-specific `client.id`s once the sidecar is used by other IDEs than VS Code. I feel it's important to include the IDE name so that users can link the `client.id` to their usage of VS Code (or another IDE).

> Also, do we want to include the sidecar version here, or the IDE's version? Do we need some unique portion of the client?

This would require the IDE to pass their version to the sidecar (or overwrite this config). I feel it's okay to keep including the sidecar version instead of the extension version for the time being, especially given that we're frequently updating the sidecar version used by the VS Code extension.

> Do we need some unique portion of the client? What about including the device ID?

Good question. I don't think this would provide a lot of value to the user. Given that including the device ID would require us to set the `client.id` within the code, I'm leaning towards not doing so. WDYT?

rhauch: > Good point. We should use IDE-specific client.ids once the sidecar is used by other IDEs than VS Code. I feel it's important to include the IDE name so that users can link the client.id to their usage of VS Code (or another IDE).

Even if we try to support running in Cursor?

> This would require the IDE to pass their version to the sidecar (or overwrite this config). I feel it's okay to keep including the sidecar version instead of the extension version for the time being, especially given that we're frequently updating the sidecar version used by the VS Code extension.

The `SidecarInfo` bean already has this information, and it's used by the `FeatureFlags` to build the LD context. We could just reuse that if we build the `client.id` in code rather than from a static template in the yaml file.

rhauch: > Good question. I don't think this would provide a lot of value to the user. Given that including the device ID would require us to set the client.id within the code, I'm leaning towards not doing so. WDYT?

My only concern is consumers, but I guess we'd use different `group.id` values if we ever started to use consumer groups (unlikely). So I guess it's fine to not include a unique value per client instance.

flippingbits: > Even if we try to support running in Cursor?

rhauch: No, but let's make that change and keep the existing `SidecarInfo` mechanism as-is

rhauch (approved):
LGTM, for now. I would like us to complete the SidecarInfo usage and use that for more dynamic client IDs. See [this comment](https://github.com/confluentinc/ide-sidecar/pull/167#discussion_r1846755548).

src/main/resources/application.yml
flippingbits: Thanks for your feedback! Are you suggesting to (1) set the env var `VSCODE_EXTENSION_VERSION` on the extension side or (2) add the config properties `ide-sidecar.ide.name` and `ide-sidecar.ide.plugin-version`? I'm fine with both because they'll lead to the same results for now. The latter will be more useful once the sidecar is used by multiple IDEs.

rhauch: I'd prefer to stick to the variables used by the `SidecarInfo` and how it's currently structured (VS Code specific props/variables, rather than generic `ide` ones), but I guess I'm open to suggestions.

flippingbits: Sounds good to me.

src/main/resources/application.yml
```diff
@@ -60,6 +60,10 @@
       # How often to check?
       interval-seconds: 60
 
+vscode:
+  extension:
+    version: unknown
+  version: unknown
```
flippingbits, line 66: These configs are used by the `SidecarInfo`. We define defaults here, so that we can use them in the `client.id` properties of the Admin, Consumer, and Producer client.

src/main/resources/application.yml
flippingbits: I updated the PR so that it uses the same variables as the `SidecarInfo`. May I ask you to take another look?

If we go down that road, I'll create an issue in the vscode repo capturing the work to be done for filling these variables with the actual versions.

rhauch (approved):
LGTM. Thanks!

PR #169, opened 2024-11-15
Populate the status for direct connections and ensure Kafka and SR proxies use credentials in headers

### LINKED ISSUE #123: Update connection `status` to work for local, CCloud and direct connections

rhauch:
Resolves #123

## Summary of Changes

Completes the functionality of direct connections, though we will add more integration tests as follow-ons. For example, the existing ITs use `LocalTestEnvironment` that runs containers without authN, and the recently-added ITs for direct connections use no credentials. We'll need another `ConfluentPlatformTestEnvironment` that runs CP clusters with several authN mechanisms enabled, and then we'll run additional tests against that.

This PR modifies the `DirectConnectionState` class to properly populate the status for direct connections, after verifying the ability to connect to Kafka using an AdminClient and to SR using an SR client.

It also fixes a few issues with how we compute the authentication-related headers using the credentials. These headers are used in the Kafka REST proxy and SR proxy implementations.

It adds some debug log messages to the ‘KafkaProducerClients’, `KafkaConsumerFactory`, `SchemaRegistryClients` and `AdminClients` beans, including logging at debug level the (redacted) configuration properties for these clients. To make it easier to log the redacted configurations, the `ClientConfigurator` class was changed to return a `Configuration` object rather than a `Map<String, Object>`. The `Configuration` object has one method to get the configuration properties as a map, while the `toString()` and other methods ensure that only the redacted form of the configuration is otherwise exposed.

Finally, fixed a bug in `BasicCredentials` and `ApiKeyAndSecret` classes in the `httpClientHeaders()` implementations. This method returns the authorization-related HTTP headers, and the base64-encoded value was not being created correctly. A few unit tests were added, using obviously-fake secret values.

## Pull request checklist

- Tests:
    - [x] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/application/ReflectionConfiguration.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/application/ReflectionConfiguration.java
@@ -6,8 +6,17 @@ import com.sun.security.auth.module.KeyStoreLoginModule;
 import io.confluent.idesidecar.scaffolding.models.TemplateManifest;
 import io.confluent.kafka.schemaregistry.avro.AvroSchema;
 import io.confluent.kafka.schemaregistry.client.rest.entities.ErrorMessage;
+import io.confluent.kafka.schemaregistry.client.rest.entities.ExtendedSchema;
+import io.confluent.kafka.schemaregistry.client.rest.entities.Mode;
+import io.confluent.kafka.schemaregistry.client.rest.entities.Rule;
+import io.confluent.kafka.schemaregistry.client.rest.entities.RuleKind;
+import io.confluent.kafka.schemaregistry.client.rest.entities.RuleMode;
+import io.confluent.kafka.schemaregistry.client.rest.entities.RuleSet;
 import io.confluent.kafka.schemaregistry.client.rest.entities.Schema;
+import io.confluent.kafka.schemaregistry.client.rest.entities.SchemaEntity;
+import io.confluent.kafka.schemaregistry.client.rest.entities.SchemaRegistryServerVersion;
 import io.confluent.kafka.schemaregistry.client.rest.entities.SchemaString;
+import io.confluent.kafka.schemaregistry.client.rest.entities.SchemaTags;
 import io.confluent.kafka.schemaregistry.client.rest.entities.SchemaTypeConverter;
 import io.confluent.kafka.schemaregistry.client.rest.entities.requests.RegisterSchemaRequest;
 import io.confluent.kafka.schemaregistry.json.JsonSchema;
@@ -82,7 +91,18 @@ import org.apache.kafka.common.serialization.ByteArrayDeserializer;
         OAuthBearerUnsecuredValidatorCallbackHandler.class,
         OAuthBearerUnsecuredLoginCallbackHandler.class,
         SaslClientCallbackHandler.class,
-        ScramServerCallbackHandler.class
+        ScramServerCallbackHandler.class,
+        // Schema Registry client classes that are not registered in
+        // https://github.com/quarkusio/quarkus/blob/3.16.3/extensions/schema-registry/confluent/common/deployment/src/main/java/io/quarkus/confluent/registry/common/ConfluentRegistryClientProcessor.java
+        Mode.class,
+        ExtendedSchema.class,
+        Rule.class,
+        RuleKind.class,
+        RuleMode.class,
+        RuleSet.class,
+        SchemaEntity.class,
+        SchemaTags.class,
+        SchemaRegistryServerVersion.class,
     }
 )
 public class ReflectionConfiguration {
--- a/src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
@@ -1,6 +1,7 @@
 package io.confluent.idesidecar.restapi.cache;
 
 import com.github.benmanes.caffeine.cache.CaffeineSpec;
+import io.quarkus.logging.Log;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import org.apache.kafka.clients.admin.AdminClient;
@@ -34,9 +35,15 @@ public class AdminClients extends Clients<AdminClient> {
         clusterId,
         () -> {
           // Generate the Kafka admin client configuration
-          var config = configurator.getAdminClientConfig(connectionId, clusterId, false);
+          var config = configurator.getAdminClientConfig(connectionId, clusterId);
+          Log.debugf(
+              "Creating schema registry client for connection %s and cluster %s with configuration:\n  %s",
+              connectionId,
+              clusterId,
+              config
+          );
           // Create the admin client
-          return AdminClient.create(config);
+          return AdminClient.create(config.asMap());
         }
     );
   }
--- a/src/main/java/io/confluent/idesidecar/restapi/cache/ClientConfigurator.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/cache/ClientConfigurator.java
@@ -12,10 +12,12 @@ import io.confluent.idesidecar.restapi.exceptions.ConnectionNotFoundException;
 import io.confluent.idesidecar.restapi.models.graph.KafkaCluster;
 import io.confluent.idesidecar.restapi.models.graph.SchemaRegistry;
 import io.confluent.idesidecar.restapi.util.CCloud;
+import io.quarkus.logging.Log;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.function.Supplier;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 
 @ApplicationScoped
@@ -32,16 +34,20 @@ public class ClientConfigurator {
    * are not included.
    *
    * @param connection        the connection
-   * @param cluster           the Kafka cluster to use
-   * @param sr                the Schema Registry to use, or null if not needed
+   * @param clusterId         the ID of the Kafka cluster to use
+   * @param bootstrapServers  the bootstrap servers of the Kafka cluster to use
+   * @param srId              the ID of the Schema Registry to use, or null if not needed
+   * @param srUri             the URI of the Schema Registry to use, or null if not needed
    * @param redact            whether to redact sensitive properties
    * @param defaultProperties the default properties to use; may be overridden by computed values
    * @return the Kafka client configuration properties
    */
-  static Map<String, Object> getKafkaClientConfig(
+  public static Map<String, Object> getKafkaClientConfig(
       ConnectionState connection,
-      KafkaCluster cluster,
-      SchemaRegistry sr,
+      String clusterId,
+      String bootstrapServers,
+      String srId,
+      String srUri,
       boolean redact,
       Map<String, String> defaultProperties
   ) {
@@ -50,19 +56,19 @@ public class ClientConfigurator {
     var props = new LinkedHashMap<String, Object>(defaultProperties);
 
     // First set the bootstrap servers
-    props.put("bootstrap.servers", cluster.bootstrapServers());
+    props.put("bootstrap.servers", bootstrapServers);
 
     // Second, add any connection properties for Kafka cluster credentials (if defined)
-    var options = connection.getKafkaConnectionOptions(cluster.id()).withRedact(redact);
+    var options = connection.getKafkaConnectionOptions(clusterId).withRedact(redact);
     connection
-        .getKafkaCredentials(cluster.id())
+        .getKafkaCredentials(clusterId)
         .flatMap(creds -> creds.kafkaClientProperties(options))
         .ifPresent(props::putAll);
 
     // Add any auth properties for Schema Registry to the Kafka client config,
     // with the "schema.registry." prefix (unless the property already starts with that)
-    if (sr != null) {
-      var additional = getSchemaRegistryClientConfig(connection, sr, redact);
+    if (srUri != null) {
+      var additional = getSchemaRegistryClientConfig(connection, srId, srUri, redact);
       additional.forEach((k, v) -> {
         if (k.startsWith("schema.registry.")) {
           props.put(k, v);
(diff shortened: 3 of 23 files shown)


src/main/java/io/confluent/idesidecar/restapi/application/ReflectionConfiguration.java
```diff
@@ -82,7 +91,18 @@
...
+        // https://github.com/quarkusio/quarkus/blob/3.16.3/extensions/schema-registry/confluent/common/deployment/src/main/java/io/quarkus/confluent/registry/common/ConfluentRegistryClientProcessor.java
+        Mode.class,
+        ExtendedSchema.class,
+        Rule.class,
+        RuleKind.class,
+        RuleMode.class,
+        RuleSet.class,
+        SchemaEntity.class,
+        SchemaTags.class,
+        SchemaRegistryServerVersion.class,
```
rhauch, line 105: The Quarkus library for Confluent SR does not register all of the classes we will (or might) need at runtime, so we do this here. Most of these are related to data contracts, which were a more recent additional to SR.

src/main/java/io/confluent/idesidecar/restapi/cache/AdminClients.java
```diff
@@ -34,9 +35,15 @@ public AdminClient getClient(String connectionId, String clusterId) {
...
+          var config = configurator.getAdminClientConfig(connectionId, clusterId);
+          Log.debugf(
+              "Creating schema registry client for connection %s and cluster %s with configuration:\n  %s",
+              connectionId,
+              clusterId,
+              config
+          );
           // Create the admin client
-          return AdminClient.create(config);
+          return AdminClient.create(config.asMap());
```
rhauch, line 46: Changing the `ClientConfigurator.get*(...)` methods to return a `Configuration` object makes it much easier to get the configuration once and then use the redacted form in log message and pass the non-redacted form to the client.

src/main/java/io/confluent/idesidecar/restapi/models/graph/ConfluentRestClient.java
```diff
@@ -220,21 +220,20 @@ protected record LastRequest(String url, String json) {
...
+   * @param headers        the headers to use for the request
    * @param firstUrl       the URL of the first page of Confluent API resources
    * @param limits         the page limits; may be null if there are no limits
    * @param responseParser the parser for the items on the page
    * @param <T>            the type of items being returned
    * @return the stream of items
    */
   protected <T> Multi<T> listItems(
-      String connectionId,
+      MultiMap headers,
```
rhauch, line 231: The places where this method is used knows whether it's using the REST client for Kafka or SR. Since each type of cluster need different credentials and therefore authN-related headers, it's easier for the caller to compute the appropriate headers and simply pass them into this method.

src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/processors/ClusterStrategyProcessor.java
```diff
@@ -59,16 +59,19 @@ public Future<ClusterProxyContext> process(ClusterProxyContext context) {
...
+        case LOCAL -> confluentLocalKafkaClusterStrategy;
+        case DIRECT -> directKafkaClusterStrategy;
+        case PLATFORM -> null;
+      };
+      case SCHEMA_REGISTRY -> switch (connectionType) {
+        case CCLOUD -> confluentCloudSchemaRegistryClusterStrategy;
+        case LOCAL -> confluentLocalSchemaRegistryClusterStrategy;
+        case DIRECT -> directSchemaRegistryClusterStrategy;
+        case PLATFORM -> null;
+      };
```
rhauch, line 74: This just refactors the logic to be a bit easier to debug (and set breakpoints) and read.

src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/strategy/DirectSchemaRegistryClusterStrategy.java
```diff
@@ -1,11 +1,39 @@
...
+        var credentials = srConfig.credentials();
+        if (credentials != null) {
+          credentials.httpClientHeaders().ifPresent(map -> map.forEach(headers::add));
+        }
+      }
+    }
+    headers.add(TARGET_SR_CLUSTER_HEADER, context.getClusterId());
+
+    return headers;
+  }
```
rhauch, line 37: This is pretty much the same as `CCloudSchemaRegistryClusterStrategy` but the credentials are accessed differently.

src/main/java/io/confluent/idesidecar/restapi/credentials/ApiKeyAndSecret.java
```diff
@@ -106,7 +106,7 @@ public Optional<MultiMap> httpClientHeaders() {
     var headers = MultiMap.caseInsensitiveMultiMap();
     if (key != null || secret != null) {
       // base64 encode the username and password
-      var value = "%s:%s".formatted(key, secret.asCharArray());
+      var value = "%s:%s".formatted(key, secret.asString());
```
rhauch, line 109: This bug is super subtle: the toString() of the char array is not the same as a String created from the char array. Added a unit test for this.

src/main/java/io/confluent/idesidecar/restapi/credentials/Redactable.java
```diff
@@ -68,7 +68,7 @@ public final char[] asCharArray() {
    */
   @JsonIgnore
   public final String asString() {
-    return new String(this.asCharArray());
+    return new String(raw);
```
rhauch, line 71: The `this.asCharArray()` makes a copy of the raw array; then String makes another copy. This change avoids the extra copy.

rhauch (commented):
A few comments to help reviewers.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -48,5 +78,126 @@ public Optional<Credentials> getSchemaRegistryCredentials(String clusterId) {
...
+              .builder()
+              .state(ConnectedState.NONE)
+              .build()
+      );
+    }
+    // There is a Kafka configuration, so validate the connection by creating an AdminClient
+    // and describing the cluster.
+    try (var adminClient = createAdminClient(kafkaConfig)) {
+      var clusterDesc = adminClient.describeCluster();
+      var actualClusterId = clusterDesc.clusterId().get(5, TimeUnit.SECONDS);
```
flippingbits, line 112: Should we make such timeouts configurable via the [application.yml](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/resources/application.yml) or env vars? It would allow the extension to overwrite/change them.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -48,5 +78,126 @@ public Optional<Credentials> getSchemaRegistryCredentials(String clusterId) {
...
+      return Future.succeededFuture(
+          ConnectionStatusSchemaRegistryStatusBuilder
+              .builder()
+              .state(ConnectedState.NONE)
+              .build()
+      );
+    }
+    // There is a Schema Registry configuration, so validate the connection by creating a
+    // SchemaRegistryClient and getting the mode.
+    try (var srClient = createSchemaRegistryClient(schemaRegistryConfig)) {
```
flippingbits, line 149: Do you know if `createSchemaRegistryClient` (or the constructor of `CachedSchemaRegistryClient`) uses a timeout when creating the SR client? If not, we'd probably wait for a long time when running into network issues so we want to set a timeout.

Alternatively, we could add timeouts to the futures in lines 84 and 85, and let them fail if they exceed the timeout. If we go down that road, we'd have to build the `FAILED` state in the method `getConnectionStatus()`.

I'm leaning towards the latter option. What's your take on it?

src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/strategy/DirectSchemaRegistryClusterStrategy.java
```diff
@@ -1,11 +1,39 @@
...
+    var headers = super.constructProxyHeaders(context);
+    if (context.getConnectionState() instanceof DirectConnectionState directConnectionState) {
+      var srConfig = directConnectionState.getSpec().schemaRegistryConfig();
+      if (srConfig != null) {
+        var credentials = srConfig.credentials();
+        if (credentials != null) {
+          credentials.httpClientHeaders().ifPresent(map -> map.forEach(headers::add));
+        }
+      }
+    }
```
flippingbits, line 33: Should we throw an exception or log an error if `context.getConnectionState()` is not a `DirectConnectionState`?

flippingbits (commented):
Thanks, @rhauch! I have one question about the handling of timeouts when retrieving the health status of Direct Connections and a few very minor suggestions/questions. Otherwise, your PR looks good to me. 🎉

src/main/java/io/confluent/idesidecar/restapi/cache/ClientConfigurator.java
```diff
@@ -50,19 +56,19 @@ static Map<String, Object> getKafkaClientConfig(
...
-        .getKafkaCredentials(cluster.id())
+        .getKafkaCredentials(clusterId)
         .flatMap(creds -> creds.kafkaClientProperties(options))
         .ifPresent(props::putAll);
 
     // Add any auth properties for Schema Registry to the Kafka client config,
     // with the "schema.registry." prefix (unless the property already starts with that)
-    if (sr != null) {
-      var additional = getSchemaRegistryClientConfig(connection, sr, redact);
+    if (srUri != null) {
```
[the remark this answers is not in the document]
rhauch: `srId` can be null at this point. I've clarified the JavaDoc of the method called below, and removed it from the two `ConnectionState` methods.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -48,5 +78,126 @@ public Optional<Credentials> getSchemaRegistryCredentials(String clusterId) {
...
+      var actualClusterId = clusterDesc.clusterId().get(5, TimeUnit.SECONDS);
+      // Set the cluster ID
+      getSpec().kafkaClusterConfig().withId(actualClusterId);
+      return Future.succeededFuture(
+          ConnectionStatusKafkaClusterStatusBuilder
+              .builder()
+              .state(ConnectedState.SUCCESS)
+              .build()
+      );
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
rhauch: I was thinking no, because the message is passed to the status error, and because it is an expected situation if invalid credentials are called. (We don't want users thinking that because the sidecar logged an exception and stack trace that there is a bug within the sidecar causing the failure to connect.)

I could add it a debug log message, though users would never see that and it's not clear it would help us if trying to debug an auth failure.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -48,5 +78,126 @@ public Optional<Credentials> getSchemaRegistryCredentials(String clusterId) {
...
+      // The connection failed, so successfully return the failed state
+      return Future.succeededFuture(
+          ConnectionStatusKafkaClusterStatusBuilder
+              .builder()
+              .state(ConnectedState.FAILED)
+              .errors(
+                  new AuthErrors().withSignIn(
```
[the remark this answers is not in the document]
rhauch: The whole point of #162 is to deprecate and move away from the `status.authentication` object, since it only applies to CCloud. The CCloud-related status is available under a nested `status.ccloud` object (structurally identical to `status.authentication`), while direct connection errors (incl auth-related errors) are recorded under `status.kafka_cluster` or `status.schema_registry` (depending upon where the error is).

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
rhauch: I don't anticipate the extension needing to change this, but I can add a configuration. How about a `ide-sidecar.connections.direct.timeout-seconds` property?

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
rhauch: The [default timeout](https://docs.confluent.io/platform/current/clients/confluent-kafka-dotnet/_site/api/Confluent.SchemaRegistry.CachedSchemaRegistryClient.html#Confluent_SchemaRegistry_CachedSchemaRegistryClient_DefaultTimeout) for all REST API calls made by the `CachedSchemaRegistryClient` is 30 seconds. It's configurable, so I can just reuse the same timeout value constant [discussed above](https://github.com/confluentinc/ide-sidecar/pull/169#discussion_r1846123371).

src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/strategy/DirectSchemaRegistryClusterStrategy.java
rhauch: We're not doing that in [ConfluentCloudSchemaRegistryClusterStrategy](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/strategy/ConfluentCloudSchemaRegistryClusterStrategy.java), and I expect that's the case because [ClusterStrategyProcessor](https://github.com/confluentinc/ide-sidecar/pull/169/files#diff-9783c6e6d6f537d0c569f437e05cc5227a5e8944eccb79842ae963104f036cb9) is always using it correctly and therefore it should never occur at runtime.

That means that throwing an exception will only help during development, but maybe that's reason enough.

rhauch: Actually, if we want to add that, I think we should do that in a followup, since we'd also want to add it in multiple other places.

flippingbits: Doing it in a follow-up sounds good to me. I wanted to make sure that we don't swallow any errors.

rhauch:
Thanks for the review, @flippingbits. I've incorporated your feedback -- would you please take a look?

flippingbits (approved):
Thanks for the update, @rhauch. Your PR looks good to me.

Issue #161, opened 2024-11-14
Add integration tests for direct connections


rhauch:
We added direct connections (without authN) in #122, basic credentials and API key+secret credentials in #124, and updated the connection statuses in #123. These PRs relied mostly upon manual testing by connecting to an LKC+SR in CCloud, and so this is the bare minimum.

We need to add integration tests to verify that direct connections work as expected with basic credentials. These can build upon the recent changes in #134 that added new `*WithDirect` IT subclasses just without authN. We'll need to change `LocalTestEnvironment` (or subclass?) to add a CP container for Kafka (w/ multiple auth mechanisms enabled) and another for CP SR (w/ basic authN enabled).

rhauch:
Previously, our integration tests all extended the `AbstractSidecarIT` class, which used the [TestContainers “singleton pattern”](https://testcontainers.com/guides/testcontainers-container-lifecycle/#_using_singleton_containers) to ensure that the `LocalTestEnvironment`’s two containers (Confluent Local container and an SR container) are started before any of the test methods and is available for reuse by all subclasses. The Ryuk container is then responsible for shutting down the containers.
And, to test different connections, we needed to subclass the `AbstracdtSidecarIT` subclasses.

The inheritance hierarchy basic looked like:
* `AbstractSidecarIT` -- defined the `LocalTestEnvironment` and started the containers before all tests and did not shut them down
  * `<FunctionalArea>IT` -- abstract class that defines the test methods for the functional area that run against the _current_ connection (which is not yet defined)
    * `<FunctionalArea>LocalIT` -- set up the current local connection before all inherited test methods are run
    * `<FunctionalArea>DirectIT` -- set up the current direct connection before all inherited test methods are run

This approach had a few benefits:
1. The containers in the `LocalTestEnvironment`, the Confluent Local container and SR container, were started once, used for all integration tests, and stopped
2. We could run the same tests against different types of connections to the `LocalTestEnvironment` (e.g., the `*LocalIT` and `*DirectIT` subclasses)
3. A set of tests are run against a single connection, reducing the setup time and overhead of the tests.

However, this also had a several limitations:
1. This resulted in a lot of IT test classes, sprinkled throughout the test source packages.
2. It is impossible to run the same tests against other `TestEnvironment` instances, such as one that runs CP containers, unless all `TestEnvironment` instances and their containers were all run at once.

The second limitation is actually the big problem, and would result in a lot of duplicated code or running fewer tests against the different environments.

[...]

PR #175, opened 2024-11-20
Refactored integration tests so that we can easily run tests against multiple TestEnvironment

rhauch:
See #161

## Summary of Changes

Previously, our integration tests all extended the `AbstractSidecarIT` class, which used the [TestContainers “singleton pattern”](https://testcontainers.com/guides/testcontainers-container-lifecycle/#_using_singleton_containers) to ensure that the `LocalTestEnvironment`’s two containers (Confluent Local container and an SR container) are started before any of the test methods and is available for reuse by all subclasses. The Ryuk container is then responsible for shutting down the containers.
And, to test different connections, we needed to subclass the `AbstracdtSidecarIT` subclasses.

This approach had a few benefits:
1. The containers in the `LocalTestEnvironment`, the Confluent Local container and SR container, were started once, used for all integration tests, and stopped
2. We could run the same tests against different types of connections to the `LocalTestEnvironment`.
3. A set of tests are run against a single connection, reducing the setup time and overhead of the tests.

However, this also had a several limitations:
1. This resulted in a lot of IT test classes.
2. It is impossible to run the same tests against other `TestEnvironment` instances, such as one that runs CP containers, unless all `TestEnvironment` instances and their containers were all run at once.

This refactors the way we define integrations tests:
* The actual integration test classes are defined together, in a new `integration` package in the test source code.
* Test methods are defined as default methods on _interfaces_, so that multiple IT test classes can run the tests by simply implementing one or more of these _IT suite_ interfaces.
* All tests that use the `LocalTestEnvironment` are defined in the `LocalITs` test class. The containers in this environment are started before the tests are run, and shut down after the tests are run.
* All tests that use the forthcoming CP `TestEnvironment` will be defined in the `ConfluentPlatformITs` test class. The containers in this environment are started before the tests are run, and shut down after the tests are run.
* We can easily add tests against other `TestEnvironments`.
* We can choose to reuse (some or all of) the test methods in future system tests.
* Many tests are run against a single connection, reducing the setup time and overhead of the tests.

One minor change that impacts production (non-test) code: the `ClientConfigurator` static methods took a Kafka cluster ID and SR ID as input parameters, though they were unused. A few of our ITs create `SimpleConsumer` instances, and for various reasons the unused IDs were fabricated unnecessarily. So the unused IDs were removed from the `ClientConfigurator` methods and the places (production and test code) where they are called.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/test/java/io/confluent/idesidecar/restapi/integration/AbstractIT.java
```diff
@@ -0,0 +1,358 @@
...
+ * {@link TestEnvironment#directConnectionSpec()}.
+ *
+ * <p>Other tests may want to use a different connection for each test. If the test class is
+ * testing the Connections API, then the test class may want each <i>test method</i> to create
+ * its own connection(s) via the {@link #setupConnection(String, ConnectionSpec)} methods,
+ * verify the result of the creation, and when neceesary delete the connection.
+ * The {@code @AfterEach} method should still call {@link #deleteAllConnections()} to clean up all
+ * connections that remain at the end of each test, giving a clean slate to the next test method.
+ */
+public abstract class AbstractIT extends SidecarClient implements ITSuite {
```
rhauch, line 67: This is mostly from the `AbstractSidecarIT` that was renamed and relocated. But the `setupLocalConnection(...)` and `setupDirectConnection(...)` methods in `AbstractSidecarID` were replaced by one set of generic `setupConnection(...)` methods in this class that take the `TestEnvironment` and `ConnectionSpec` objects, various forms of which can be obtained from the `TestEnvironment` object. This makes it a bit easier to write the many different subclasses that each setup one connection.

src/test/java/io/confluent/idesidecar/restapi/kafkarest/api/ClusterV3Suite.java
```diff
@@ -2,14 +2,14 @@
 
 import static org.hamcrest.Matchers.equalTo;
 
-import io.confluent.idesidecar.restapi.util.AbstractSidecarIT;
+import io.confluent.idesidecar.restapi.integration.ITSuite;
 import io.confluent.idesidecar.restapi.util.ConfluentLocalKafkaWithRestProxyContainer;
 import org.junit.jupiter.api.Test;
 
-abstract class ClusterV3ApiImplIT extends AbstractSidecarIT {
+public interface ClusterV3Suite extends ITSuite {
```
rhauch, line 9: This is an example of one of these `*IT` classes that was extended by a `*LocalIT` class and a `*DirectIT` class. The `*IT` class had all of the test methods, so the class was changed as follows:
* The `*IT` abstract base class was made an interface that extends `ITSuite` rather than `AbstractSidecarIT` -- these are runnable sets of test classes now, and instead are merely interfaces that define test methods.
* The `@Test` methods were turned into default methods that depend only on the methods (some abstract) inherited from `ITSuite`

rhauch: The `LocalITs` and `ConfluentPlatformITs` classes now contain the test classes which get the test methods from one or more of these "suite" interfaces.

src/test/java/io/confluent/idesidecar/restapi/integration/LocalITs.java
```diff
@@ -0,0 +1,267 @@
...
+  static void beforeAll() {
+    // Start the test environment and use it for all tests
+    TEST_ENVIRONMENT.start();
+  }
+
+  @AfterAll
+  static void afterAll() {
+    // Shutdown the test environment after all tests have run
+    TEST_ENVIRONMENT.shutdown();
+  }
```
rhauch, line 36: All tests defined within this `LocalITs` test class use the `LocalTestEnvironment`, which is started before all tests and stopped after all tests.

src/test/java/io/confluent/idesidecar/restapi/integration/LocalITs.java
```diff
@@ -0,0 +1,267 @@
...
+
+  @Nested
+  class LocalConnectionTests {
```
rhauch, line 39: A `@Nested` test class exists for each kind of connection to be tested. This is for local connections. This could just implement all of the IT suite interfaces, but instead we define them on multiple `@Nested` test classes in this test class.

src/test/java/io/confluent/idesidecar/restapi/integration/LocalITs.java
```diff
@@ -0,0 +1,267 @@
...
+      public TestEnvironment environment() {
+        return TEST_ENVIRONMENT;
+      }
+
+      @BeforeEach
+      @Override
+      public void setupConnection() {
+        setupConnection(this, TestEnvironment::localConnectionSpec);
+      }
+    }
```
rhauch, line 93: A `@Nested` test class exists for each kind of connection to be tested: the `LocalConnectionTests` here is for local connections. This class could have implemented all of the IT suite interfaces, but that would have put a hundred tests (including dynamic tests) all on this one class -- the test results were a bit hard to navigate.

So instead, we define multiple `@Nested` (2-levels deep) test classes that implement a subset of the _IT suite_ interfaces. This organizes them a bit more nicely. For example, above the `RecordTests` implements the `RecordsV3ApiSuite` interface and thus gets all of its test methods and runs them against a _local_ connection that is reused for all of the tests in `LocalConnectionTests`.

Likewise, there is a `RecordFailuresTest` that implements the `RecordsV3ApiErrorsSuite` tests, a `TopicsTest` that implements the `TopicV3Suite` and `PartitionV3Suite` tests, etc.

Lower down in this file is a top-level `@Nested` test class called `DirectConnectionWithoutCredentialsTests`, with nested test classes that implement various test suites, that runs *the same tests* against a direct connection with no credentials using the `LocalTestEnvironment`'s containers.

rhauch:
I've added [a comment](https://github.com/confluentinc/ide-sidecar/issues/161#issuecomment-2489147944) to #161 that describes the new design of these integration tests.

src/test/java/io/confluent/idesidecar/restapi/integration/AbstractIT.java
```diff
@@ -0,0 +1,348 @@
...
+      var spec = connectionSpec;
+      spec = spec.withName( "%s (%s)".formatted(spec.name(), testScope));
+      spec = spec.withId( "%s-%s".formatted(spec.id(), testScope));
+
+      // Create the connection we'll use
+      createConnection(spec);
+
+      // Get the clusters we'll use
+      var kafkaCluster = getKafkaCluster().orElseThrow();
+      var srCluster = getSchemaRegistryCluster().orElseThrow();
```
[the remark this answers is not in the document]
rhauch: Yeah, those would be good improvements to make. WDYT about making them as part of #166? I think we'd want to change `orElseThrow()` here to `orElse(null)` and then handle the case where the `ScopedConnection.srCluster` field is used to make sure we're properly handling the case. Maybe the same thing while we're here to handle the case when there is no Kafka cluster and just SR?

As you suggested offline, we could then _in the future_ define an Apache Kafka only `TestEnvironment` that does not have an SR, and we could run the test suites that don't use schemas against AK.

rohitsanj (approved):
LGTM. Thanks again for spending the time on making our integration test suite super easy to work with and extend into setting up system tests later. Exciting stuff. Will pay dividends very soon.

PR #178, opened 2024-11-20
add schema errors class

Cerchie:
This PR creates a cache of caches for SchemaError clearing, reading, and writing, to be used by instances of MessageViewerClusterInfoProcessor(s). 

Questions:

This class creates a key:value pair of connectionId:clusterId -- is this the right way to think about this? Or should it be like schemaId and be handled separately. 

TBD

## Pull request checklist

- Tests:
    - [ ] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


flippingbits:
> This class creates a key:value pair of connectionId:clusterId -- is this the right way to think about this? Or should it be like schemaId and be handled separately.

Great question! I think it makes sense to use the same approach for schema-related errors because of RBAC; multiple connections might reference the same schemaId but use different credentials/principals to authenticate. This would also allow us to clear the cache when a connection gets updated/deleted ([see how the ide-sidecar implements this for the ClusterCache](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/cache/ClusterCache.java#L214-L219)).

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -0,0 +1,85 @@
+package io.confluent.idesidecar.restapi.clients;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+@ApplicationScoped
+public class SchemaErrors {
+
+  private final Map<String, Map<String, String>> cacheOfCaches = new ConcurrentHashMap<>();
```
flippingbits, line 10: WDYT of storing the `connectionId`s in the keys of the `Map` at the first level and storing the `clusterId`s/`schemaId`s in the key of the `Map` at the second level? This could simplify this class and make methods, like `generateKey` or `generateSchemaKey`, obsolete.

Cerchie: I like it!

rhauch: It would also be good to use concrete types for the first level values and maybe second level values. And we might think about using record types for the keys; definitely for the second level (since it's composite), though maybe's it's okay to just use a string for the first level as we use String Connection IDs everwhere else:

For example, maybe:
```
record ConnectionId(String id) {}
record SchemaId(String clusterId, String schemaId) {}
record Error(String message) {}
class ErrorsForConnection {
  private final ConnectionId id;
  private final Cache<SchemaId, Error> errorsBySchemaId = ...
}

private final Map<ConnectionId, ErrorsForConnection> errorsByConnectionId = ...

...
```
Nit: "connection error" might ambiguously refer to connectivity errors or errors for connections.

Using types helps in a few ways:
* Adds type safety, to avoid risking using incorrect keys or values.
* Makes the code easier to read and understand, by providing distinct terms
* If needed, we can add more to `Error`, such as maybe code, status, etc.
* Instances of record types are immutable
* We can add custom methods, if needed.

rhauch: As alluded to above, we probably want to use a Caffeine `Cache` so that the individual `Error` entries will be purged after they are added; e.g., `expireAfterWrite(Duration)`.
```
private final Cache<SchemaId, Error> errorsBySchemaId = Caffeine.newBuilder()
      .expireAfterWrite(SCHEMA_FETCH_ERROR_TTL)
      .build();
```

And we might want to actually use a Cache at the top level, so that the entry associated with a connection will be purged some time after it is last used; e.g., `expireAfterAccess(Duration)`:
```
private final Cache<ConnectionId, ErrorsForConnection> errorsByConnectionId = Caffeine.newBuilder()
      .expireAfterAccess(CONNECTION_TTL)
      .build();
```
See my next comment about handling connection lifecycle events.

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -0,0 +1,85 @@
...
+    getSubCache(generateKey(connectionId, clusterId)).put(clusterId, error);
+  }
+
+  public void clearByConnectionId(String connectionId, String clusterId) {
+    Map<String, String> subCache = getSubCache(generateKey(connectionId, clusterId));
+    subCache.remove(clusterId);
+    if (subCache.isEmpty()) {
+      cacheOfCaches.remove(generateKey(connectionId, clusterId));
+    }
+  }
```
rhauch, line 39: I wonder if we should have a method that response to connection events, like on `ClusterCache`, when the connection is closed:
```
  void onConnectionDeleted(@ObservesAsync @Lifecycle.Deleted ConnectionState connection) {
    errorsBySchemaId.invalidate(connection.getSpec().id());
  }
```
The risk is that we remove the errors when a connection is closed, but meanwhile the sidecar is still processing a previously-submitted consume request that still needs the schema failures.

On second thought, we could not have this method and instead rely upon the CONNECTION_TTL to automatically evict from the `errorsByConnectionId` cache the `ErrorsForConnection` entry for a disconnected/closed connection, some TTL after it is unused. I suspect the sizes of these caches will not be huge (home many failed fetches by schema ID might one connection have), so it's okay to keep them around for a bit longer than the connection exists.

The actual TTL depends a lot on the SCHEMA_FETCH_ERROR_TTL, which corresponds to the existing `RecordDeserializer.CACHE_FAILED_SCHEMA_ID_FETCH_DURATION` that is only 30 seconds. Basically, after 30 seconds of _non-use_, the error for a `SchemaId` will be purged anyway, which means after 30 seconds, the `ErrorsForConnection` will be empty anyway. So if `CACHE_FAILED_SCHEMA_ID_FETCH_DURATION` is 30 seconds, then maybe `CONNECTION_TTL` could be 60 seconds (since a thread processing the consume request my hold onto the `ErrorsForConnection` object for 0-5-ish seconds, with a bit of a buffer).

But, we probably **do** want to clear any `ErrorsForConnection` entry when a connection is created or updated, just in case the SR URI is changed on the connection:
```
  void onConnectionCreated(@ObservesAsync @Lifecycle.Created @Lifecycle.Updated ConnectionState connection) {
    errorsBySchemaId.invalidate(connection.getSpec().id());
  }
```

flippingbits: > The risk is that we remove the errors when a connection is closed, but meanwhile the sidecar is still processing a previously-submitted consume request that still needs the schema failures.

I'm not sure if there's a real risk. In this case, the sidecar (or the used SR client) would hit the proxied SR routes again, which will return a proxy error because the connection ID does no longer exist. So, invalidating the cache after deleting a connection could make sense because the errors will change.

> On second thought, we could not have this method and instead rely upon the CONNECTION_TTL to automatically evict from the errorsByConnectionId cache the ErrorsForConnection entry for a disconnected/closed connection, some TTL after it is unused. I suspect the sizes of these caches will not be huge (home many failed fetches by schema ID might one connection have), so it's okay to keep them around for a bit longer than the connection exists.

Good point. From my POV, cache invalidation is less about saving space but more about returning correct responses. Let's assume a user runs into errors because they have accidentally provided incorrect SR credentials. If we don't invalidate the cache when they fix the credentials by recreating or updating the connection, we'd continue serving the errors despite valid credentials being present.

> But, we probably do want to clear any ErrorsForConnection entry when a connection is created or updated

Yeah - I think, for the sake of keeping the implementation as simple and correct as possible, we should continue using the `SCHEMA_FETCH_ERROR_TTL`, not introduce the `CONNECTION_TTL`, and invalidate the cache when a connection has been created/updated/deleted. WDYT?

src/main/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializer.java
```diff
@@ -64,10 +64,7 @@ public class RecordDeserializer {
   private static final ObjectMapper AVRO_OBJECT_MAPPER = new AvroMapper(new AvroFactory());
   public static final byte MAGIC_BYTE = 0x0;
   private static final Duration CACHE_FAILED_SCHEMA_ID_FETCH_DURATION = Duration.ofSeconds(30);
-  private static final Cache<Integer, String> SCHEMA_FETCH_ERROR_CACHE = Caffeine.newBuilder()
-      .expireAfterWrite(CACHE_FAILED_SCHEMA_ID_FETCH_DURATION)
-      .build();
-
+  private static SchemaErrors schemaErrors = new SchemaErrors();
```
flippingbits, line 67: We need to `@Inject` the `SchemaErrors` because it's a bean (see [this guide on dependency injection in Quarkus](https://quarkus.io/guides/cdi)). We can look at it together when pairing later today.

flippingbits, line 67:
```suggestion
  @Inject
  private SchemaErrors schemaErrors;
```

We need to `@Inject` the `SchemaErrors` because it's a bean (see [this guide on dependency injection in Quarkus](https://quarkus.io/guides/cdi)). You can then remove the `schemaErrors` parameter from the constructor. Also, we don't need to make this variable `static` because there is only one instance of the `RecordDeserializer` (see this useful [blog post about bean scopes](https://marcelkliemannel.com/articles/2021/overview-of-bean-scopes-in-quarkus/#short-overview-of-quarkus-scopes)). We can look at it together when pairing later today.

Cerchie:
huh. tests still passing locally, and 

```
On branch cerchie/add-schema-errors-class
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   src/generated/resources/openapi.json
        modified:   src/generated/resources/openapi.yaml

no changes added to commit (use "git add" and/or "git commit -a")
```

src/test/java/io/confluent/idesidecar/restapi/resources/KafkaConsumeResourceTest.java
```diff
@@ -164,7 +164,7 @@ void testNonExistentClusterInfoReturns404() {
         .statusCode(404)
         .contentType(MediaType.APPLICATION_JSON)
         .body("title", containsString(
-            "Kafka Cluster %s not found in connection %s".formatted(KAFKA_CLUSTER_ID, CONNECTION_ID)
+            "not found in connection %s".formatted(CONNECTION_ID)
```
flippingbits, line 167: What's the reason for this change?

flippingbits (commented):
Thanks for the update, @Cerchie! 🎉 I have minor suggestions and a few questions. Otherwise, your PR looks good to me.

flippingbits (approved):
Thanks a lot for the updates and your persistence, @Cerchie! Your PR looks good to me. 🎉 

May I trouble you to take care of the remaining minor suggestions in a follow-up PR shortly?

Cerchie:
> Thanks a lot for the updates and your persistence, @Cerchie! Your PR looks good to me. 🎉
> 
> May I trouble you to take care of the remaining minor suggestions in a follow-up PR shortly?

of course! slated for first thing Monday

src/test/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializerTest.java
```diff
@@ -352,7 +363,7 @@ record TestCase(
     ) {
       static TestCase nonRetryable(int statusCode) {
         // Expect only 1 try when the status code is non-retryable
-        return new TestCase(statusCode, 3, 1, null);
+        return new TestCase(statusCode, 1, 1, null);
```
rohitsanj, line 366: Hmm, did we fix a bug in the implementation (and hence in the test)?

flippingbits: This change should not have any affect because `assertSchemaFetchFailuresCached(...)`'s parameter `maxRetries` seems to be unused. We might want to refactor this in a follow-up PR.

PR #180, opened 2024-11-21
Remove the `kafka_cluster.id` from the connection specs, and update RealDirectFetcher to populate the cluster ID

### LINKED ISSUE #173: Kafka cluster ID is not set correctly in cached Kafka cluster entries
When creating direct connections and specifying a Kafka cluster's bootstrap servers, the cluster ID specified with the connection specification must exactly match the actual ID of the cluster. This is problematic when the cluster ID is not known at the time of connection. For example, with [cp-demo](https://docs.confluent.io/platform/current/tutorials/cp-demo/overview.html), the cluster ID _can_ be specified or otherwise will be generated (and is difficult to find).

Currently, the `DirectConnectionState` uses the `AdminClient` to connect to the cluster using the supplied credentials, to obtain the cluster description and the cluster ID, and the success or failure is reported in the connection's status. Meanwhile, the `RealDirectFetcher` that is used to compute the GraphQL resources for the connection does not currently reconnect to the cluster to fetch the cluster ID, and instead sets the cluster ID within the internally cached `KafkaCluster` object to a generated ID that likely will not match the actual cluster ID.

There are several changes that need to be made:
1. The `spec.kafka_cluster.id` should be removed; users should not be required to enter this information, and it should always be fetched from the cluster itself (see next item).
2. Instead, the ID for the direct connections' `KafkaCluster` should be read from the cluster itself, using either the `AdminClient`. This would mean that the `RealDirectFetcher` would also make use of a transient `AdminClient` object to access cluster information, but this is the purpose of the fetcher.

rhauch:
Resolves #173 

## Summary of Changes

We should not allow users to specify the ID of the Kafka cluster in direct connection specs — getting the ID is not always straightforward, and even if they provide an ID it might be wrong.

Instead, the `RealDirectFetcher` should be able to use the AdminClient to get the cluster’s ID from the cluster, and then include the ID in the cached `KafkaCluster` objects.

Added integration tests for the Connection API and GraphQL API to test local and direct connections and the Kafka cluster and Schema Registry cluster representations (including the Kafka cluster does have an ID). (Unfortunately, we can’t easily mock Kafka protocol, so adding unit tests for all the `RealDirectFetcher` logic is not as straightforward.)

Found and fixed a bug where the `status` response in a newly-created direct connection had no `kafka_cluster` or `schema_registry` object.

## Pull request checklist

- Tests:
    - [x] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
@@ -15,6 +15,7 @@ import io.confluent.idesidecar.restapi.models.ConnectionStatusKafkaClusterStatus
 import io.confluent.idesidecar.restapi.models.ConnectionStatusSchemaRegistryStatusBuilder;
 import io.confluent.kafka.schemaregistry.client.CachedSchemaRegistryClient;
 import io.confluent.kafka.schemaregistry.client.SchemaRegistryClient;
+import io.quarkus.logging.Log;
 import io.smallrye.common.constraint.NotNull;
 import io.smallrye.common.constraint.Nullable;
 import io.vertx.core.Future;
@@ -25,6 +26,7 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.eclipse.microprofile.config.ConfigProvider;
 
@@ -41,6 +43,20 @@ public class DirectConnectionState extends ConnectionState {
           .orElse(5L)
   );
 
+  static final ConnectionStatus INITIAL_STATUS = new ConnectionStatus(
+      null,
+      new KafkaClusterStatus(
+          ConnectedState.ATTEMPTING,
+          null,
+          null
+      ),
+      new SchemaRegistryStatus(
+          ConnectedState.ATTEMPTING,
+          null,
+          null
+      )
+  );
+
   public DirectConnectionState() {
     super(null, null);
   }
@@ -50,6 +66,11 @@ public class DirectConnectionState extends ConnectionState {
       @Nullable StateChangedListener listener
   ) {
     super(spec, listener);
+  }
+
+  @Override
+  public ConnectionStatus getInitialStatus() {
+    return INITIAL_STATUS;
   }
 
   public MultiMap getAuthenticationHeaders(ClusterType clusterType) {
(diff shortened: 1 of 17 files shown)


src/test/java/io/confluent/idesidecar/restapi/util/SidecarClient.java
```diff
@@ -691,13 +691,34 @@ public ValidatableResponse submitLocalConnectionsGraphQL() {
 
   public ValidatableResponse submitDirectConnectionsGraphQL() {
     return submitGraphQL(
-        loadLocalConnectionsGraphQL()
+        loadDirectConnectionsGraphQL()
```
rhauch, line 694: Found and fixed this bug in the IT infra.

src/test/java/io/confluent/idesidecar/restapi/util/SidecarClient.java
```diff
@@ -691,13 +691,34 @@ public ValidatableResponse submitLocalConnectionsGraphQL() {
...
     );
   }
 
   public ValidatableResponse submitCCloudConnectionsGraphQL() {
     return submitGraphQL(
-        loadLocalConnectionsGraphQL()
+        loadCCloudConnectionsGraphQL()
```
rhauch, line 700: Found and fixed this bug in the IT infra.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -105,76 +126,105 @@ public Future<ConnectionStatus> getConnectionStatus() {
...
+                .state(ConnectedState.FAILED)
+                .errors(
+                    new AuthErrors().withSignIn(
+                        "Failed to connect to Kafka cluster: %s".formatted(cause.getMessage())
+                    )
+                ).build()
+        );
+      }
+    }).orElseGet(() -> {
+      // There is no Kafka cluster configuration, so return the NONE state
```
rhauch, line 162: This just refactors the `DirectConnectionState` to have a public `withAdminClient(...)` method and to use that here. The `RealDirectFetcher` can then use the `withAdminClient(...)`, avoiding some duplicate code.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -105,76 +126,105 @@ public Future<ConnectionStatus> getConnectionStatus() {
...
+                .state(ConnectedState.FAILED)
+                .errors(
+                    new AuthErrors().withSignIn(
+                        "Failed to connect to Schema Registry: %s".formatted(e.getMessage())
+                    )
+                ).build()
+        );
+      }
+    }).orElseGet(() -> {
+      // There is no Schema Registry configuration, so return the NONE state
```
rhauch, line 198: This just refactors the `DirectConnectionState` to have a public `withSchemaRegistryClient(...)` method and to use that here. The `RealDirectFetcher` can then use the `withSchemaRegistryClient(...)`, avoiding some duplicate code.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealDirectFetcher.java
```diff
@@ -106,64 +106,72 @@ public List<DirectConnection> getConnections() {
...
-        );
-      }
-    }
-    if (cluster != null) {
-      onLoad(connectionId, cluster);
-      return Uni.createFrom().item(cluster);
+    var state = connections.getConnectionState(connectionId);
+    if (state instanceof DirectConnectionState directState) {
+      var spec = state.getSpec();
+      // Use the admin API to obtain the cluster ID
```
rohitsanj, line 143: This comment needs fixing -- perhaps we just remove it.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealDirectFetcher.java
```diff
@@ -106,64 +106,72 @@ public List<DirectConnection> getConnections() {
...
+          } else {
+            cluster = new DirectSchemaRegistry(
+                orDefault(srConfig.id(), () -> connectionId + "-schema-registry"),
+                srConfig.uri(),
+                connectionId
+            );
+          }
+          result = Uni.createFrom().item(cluster);
+        } catch (Throwable t) {
+          Log.debugf("Failed to create schema registry client for connection ID=%s", connectionId, t);
```
rohitsanj, line 167: Shouldn't this say something like "failed to connect"? Technically at this point we have an SR client instance that was used to make the global `/mode` API call.

```suggestion
          Log.debugf("Failed to connect to schema registry for connection ID=%s", connectionId, t);
```

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealDirectFetcher.java
```diff
@@ -106,64 +106,72 @@ public List<DirectConnection> getConnections() {
...
+          result = Uni.createFrom().nullItem();
+        }
+        return result;
+      }).orElseGet(() -> Uni.createFrom().nullItem());
```
rohitsanj, line 171: Perhaps the "failed to create client" debug log should go in this supplier function.

```suggestion
      }).orElseGet(() -> {
        Log.debugf("Failed to create schema registry client for connection ID=%s", connectionId, t);
        Uni.createFrom().nullItem();
      });
```

rohitsanj (commented):
Looks good, thanks! 

I have a few very minor comments.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -41,6 +43,20 @@ public class DirectConnectionState extends ConnectionState {
...
+          ConnectedState.ATTEMPTING,
+          null,
+          null
+      ),
+      new SchemaRegistryStatus(
+          ConnectedState.ATTEMPTING,
+          null,
+          null
+      )
+  );
```
flippingbits, line 58: Using `ConnectedState.ATTEMPTING` as the initial state of Direct connections is fine for now, given that the endpoints `GET /gateway/v1/connections` and  `GET /gateway/v1/connections/:id` call `DirectConnectionState#getKafkaConnectionStatus()`, which synchronously checks the status of the connection. 

With the availability of Direct connections, I expect users to manage more than one connection. As a consequence, I think it's no longer feasible to perform the status check synchronously when calling the endpoint `GET /gateway/v1/connections`. So, we probably want to migrate to asynchronously checking the status of Direct connections (if not of all connection types) sooner rather than later, in which case we should use `ConnectedState.NONE` as the initial state and set the state to `ConnectedState.ATTEMPTING` when attempting to establish a connection to the Kafka cluster. WDYT? I could capture it in an issue if you think it's worth addressing it.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
flippingbits, line 162: Do you think we should require a Kafka cluster config for Direct connections?

flippingbits (dismissed):
Great work, @rhauch! I have two questions that are more about follow-up/future work so you don't need to address them in this PR. I still wanted to get the discussion started.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealDirectFetcher.java
rhauch: Any _uncaught_ exception is logged as an error in the `withSchemaRegistry(...)` method, and the contract is that the function should handle all exceptions on its own.

What's interesting is that we can create instances of the `KafkaAdminClient` or the `SchemaRegistryClient` with incorrect credentials or wrong information (e.g., bootstrap servers, URL), but only when we try and use those clients will they actually try to connect by making calls over the network.

This is why the `withSchemaRegistry(...)` and `withAdminClient(...)` methods in `DirectConnectionState` create the client objects safely before calling the supplied function, **and** expect the supplied function to handle all exceptions. This means that any exception caught in the `with...` methods will be only from the supplied function, meaning it is unexpected and therefore it's a bug that needs to be fixed. And that's why such exceptions are logged as errors.

So, the `orElseGet(...)` block here is called only when there is no schema registry configuration for this connection, or if there is an uncaught exception in the function (at which point it's only logged). I can add comments and log debug messages to help make that clear, but the log message should _not_ be "failed to create a connection" in the `orElseGet(...)` block.

rhauch: Actually, my statement above about always being able to create a KafkaAdminClient is not quite correct: credentials are indeed only checked once you do something with the client, but the bootstrap server addresses are resolved immediately upon creation of the KafkaAdminClient, so I'll handle that in this PR.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
rhauch: Great suggestion. And thanks for capturing an issue, when you get a chance.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
rhauch: I'd prefer to leave it as-is, where you have to have _either_ a Kafka cluster or an SR or both. Although we don't expose them yet, there are some valid use cases for connecting to an SR and just viewing or working with schemas w/o having to work with Kafka topics that use those schemas.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealDirectFetcher.java
rhauch: Improved the error messages.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealDirectFetcher.java
rhauch: Improved how exceptions are handled, making it much easier to handle all errors in one spot.

rohitsanj (approved):
LGTM!

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
flippingbits: Thanks! I created https://github.com/confluentinc/ide-sidecar/issues/181.

PR #184, opened 2024-11-22
Support Connection API for testing possible connections

### LINKED ISSUE #179: Allow testing a new connection with a "dry run" type of request
Implement an endpoint that `vscode` can use before creating a new connection that would test the connection is valid instead of saving it. It would be used in the New Connection form UI as a "Test Connection" button (or similar), which is common in other connection form UIs

rhauch:
Resolves #179

## Summary of Changes

Expands the `POST /gateway/v1/connections` endpoint to add a `dry_run` query parameter that defaults to `false` for current behavior. When `dry_run=true`, the endpoint will not create a new connection but will instead return the Connection resource type response with a status that contains the information _as if_ the connection were created. This will allow the extension to have a “Test” button on the new connection panel, without having to actually create, wait for updated status, and then delete the connection.

The behavior of the existing endpoint will be changed as follows:
* When the `dry_run` query parameter is `false` (the default), the endpoint calls `ConnectionStateManager.createConnectionState(spec)` that checks for duplicate IDs (if provided), validates the input spec (failing if there are validation errors), adds the new connection in the `ConnectionStateManager`, and returns a Connection object with the supplied `ConnectionSpec` and _initial_ `ConnectionStatus`.
* When the `dry_run` query parameter is `true`, the endpoint calls a new method `ConnectionStateManager.testConnection(spec)` that checks for duplicate IDs (if provided), validates the input spec (failing if there are validation errors), and *synchronously* computes the status (via `ConnectionState.getConnectionStatus()`), and finally returns a Connection object with the supplied `ConnectionSpec` and computed `ConnectionStatus`.

Found and fixed a bug in the `DirectConnectionState` that was incorrectly setting the `status.schema_registry.state` rather than null when there is no schema registry configured.

## Pull request checklist

- Tests:
    - [x] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/main/java/io/confluent/idesidecar/restapi/resources/ConnectionsResource.java
```diff
@@ -83,11 +84,35 @@ public Uni<ConnectionsList> listConnections() {
...
+      // Get the status of the connection
+      var futureStatus = testedState.getConnectionStatus();
+      // And create a uni that will complete when the status is available
+      return Uni
+          .createFrom()
+          .completionStage(futureStatus.toCompletionStage())
+          .map(connectionStatus ->
+              Connection.from(testedState, connectionStatus)
+          );
+    }
```
rhauch, line 108: This is the bulk of the changes in the Connections API to add the `dry_run` query parameter to the `POST /gateway/v1/connections` endpoint.

src/main/java/io/confluent/idesidecar/restapi/resources/ConnectionsResource.java
```diff
@@ -83,11 +84,35 @@ public Uni<ConnectionsList> listConnections() {
...
+    // Create a real connection state and return the connection using the initial status
+    var newState = connectionStateManager.createConnectionState(connectionSpec);
+    return Uni
+        .createFrom()
+        .item(
+            Connection.from(newState, newState.getInitialStatus())
```
rhauch, line 114: For real create-connection behavior, I changed this line to `Connection.from(newState)` to this, and removed the `Connection.from(ConnectionState state)` (without a status) to help ensure we always pass in the correct state.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -221,13 +218,8 @@ protected Future<SchemaRegistryStatus> getSchemaRegistryConnectionStatus() {
...
     }).orElseGet(() -> {
-      // There is no Schema Registry configuration, so return the NONE state
-      return Future.succeededFuture(
-          ConnectionStatusSchemaRegistryStatusBuilder
-              .builder()
-              .state(ConnectedState.NONE)
-              .build()
-          );
+      // There is no Schema Registry configuration, so return no status for SR
+      return Future.succeededFuture(null);
```
rhauch, line 222: This is the bug I found and fixed: when there is no `spec.schema_registry`, then we want `status.schema_registry` to also be null.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -177,13 +179,8 @@ protected Future<KafkaClusterStatus> getKafkaConnectionStatus() {
...
         () -> {
-            // There is no Kafka cluster configuration, so return the NONE state
-            return Future.succeededFuture(
-                ConnectionStatusKafkaClusterStatusBuilder
-                    .builder()
-                    .state(ConnectedState.NONE)
-                    .build()
-            );
+            // There is no Kafka cluster configuration, so return a null Kafka status
+            return Future.succeededFuture(null);
```
rhauch, line 183: And here, too: when there is no `spec.kafka_cluster`, then we want `status.kafka_cluster` to also be null.

src/test/java/io/confluent/idesidecar/restapi/integration/connection/DirectConnectionSuite.java
```diff
@@ -14,12 +15,98 @@
...
+        .body("api_version", equalTo("gateway/v1"))
+        .body("kind", equalTo("Connection"))
+        .body("metadata.self", startsWith("http://localhost:26637/gateway/v1/connections/"))
+        .body("metadata.resource_name", nullValue())
+        .body("id", equalTo(spec.id()))
+        .body("spec.id", equalTo(spec.id()))
+        .body("spec.name", equalTo(spec.name()))
+        .body("spec.type", equalTo(ConnectionType.DIRECT.name()))
+        .body("status.authentication.status", equalTo(Status.NO_TOKEN.name()))
+        .body("status.kafka_cluster.state", equalTo(ConnectedState.SUCCESS.name()));
```
rhauch, line 45: I've come to really like this style of verification of the response objects. IMO, it's very clear as to what the test is asserting.

src/main/java/io/confluent/idesidecar/restapi/resources/ConnectionsResource.java
flippingbits: Kudos for making this look straightforward 👍

flippingbits (approved):
LGTM, @rhauch! Feel free to decide if you want to incorporate the minor suggestions.

rhauch:
Fixed the build failures by merging IT fixes from `main`. New build is enqueued, and should start soon.

flippingbits (dismissed):
Thanks for the update, @rhauch. LGTM! 💯

rhauch:
It's about time the build passes. :-(

rohitsanj:
🤦

rohitsanj (approved):
Trusting @flippingbits and re-approving. (Take 2)

Issue #153, opened 2024-11-08
Cache schema fetch errors by connection ID, SR cluster ID and schema ID


rohitsanj:
From https://github.com/confluentinc/ide-sidecar/pull/150/files#r1834988784:

> The NativeConsumeStrategy [reuses the same RecordDeserializer instance in every SimpleConsumer object it creates](https://github.com/confluentinc/ide-sidecar/blob/75ee1d1f4fb42754f8ab2c5db8abf99a29085fc8/src/main/java/io/confluent/idesidecar/restapi/messageviewer/strategy/NativeConsumeStrategy.java#L52-L56), and the RecordDeserialzer is caching error messages by schema ID regardless of the SR cluster. [..] we should cache the error messages by connection ID, SR cluster ID, and schema ID. This may be simpler, and this method could clear the error cache by connection ID.

Cerchie:
This has been partially addressed by https://github.com/confluentinc/ide-sidecar/pull/150, however, in the process, an opportunity to refactor and create a schemaErrors class with read/write/clear methods was discovered. 

This way, the RecordDeserializer can use the supplied cache in schemaErrors, still keyed by Schema ID, and multiple MessageViewerClusterInfoProcessor(s) could find and reuse the same cache instances. 

Pull #158 set the stage for this change.

Cerchie:
completed after follow up PR #200

PR #196, opened 2024-11-25
Base64 decode headers from Confluent Cloud Consume API

### LINKED ISSUE #183: Decode base64 encoded header values returned from CCloud Consume API into UTF-8

rohitsanj:
## Summary of Changes

Resolves #183 

The Confluent Cloud Consume REST API returns base64 encoded header values. Instead of simply returning these as-is to vscode Message Viewer, we process the header values by base64 decoding into UTF-8 strings.

I've simply updated a few test fixtures to expect base64 decoded strings instead.

Before:
Screenshot 2024-11-25 at 2 39 03 PM

After:
Screenshot 2024-11-25 at 2 37 01 PM

## Pull request checklist

- Tests:
    - [ ] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/main/java/io/confluent/idesidecar/restapi/messageviewer/strategy/ConfluentCloudConsumeStrategy.java
```diff
@@ -222,6 +225,34 @@ private PartitionConsumeData processPartition(
...
+          String decodedValue;
+          try {
+            decodedValue = new String(
+                BASE64_DECODER.decode(header.value()), StandardCharsets.UTF_8
+            );
+          } catch (IllegalArgumentException e) {
+            // For whatever reason, we couldn't decode the Base64 encoded header value.
+            // Perhaps the API changed or the data is corrupted. Either way,
+            // we log a warning and return the raw value.
+            Log.warn("Failed to base64 decode header value from Confluent Cloud", e);
```
rhauch, line 243: I'm worried about the sidecar spewing this log message w/ stack trace for a lot of messages that all fit the same pattern. In other words, **if** this happens for one record, it's probably going to spew for every record being consumed.

Is it straightforward to count these _per request_, and maybe log it only for the first one (or few)?

rohitsanj: Hm true, I think the "only log for first one" will be easy to implement once we have #153 in place.

rohitsanj: I'm starting to think that this is probably not a failure case after all. The records may have header values that are not UTF-8 strings.

flippingbits: > I'm starting to think that this is probably not a failure case after all. The records may have header values that are not UTF-8 strings.

What about changing the log level to debug?

src/main/java/io/confluent/idesidecar/restapi/messageviewer/strategy/ConfluentCloudConsumeStrategy.java
```diff
@@ -222,6 +225,34 @@ private PartitionConsumeData processPartition(
     );
   }
 
+  private static List<SimpleConsumeMultiPartitionResponse.PartitionConsumeRecordHeader>
```
flippingbits, line 228: nit: I don't think this method needs to be `static`. OTOH, it probably doesn't hurt.

rohitsanj: Ah makes sense. I frequently use the `Extract to method` feature in IntelliJ and didn't pay attention to whether it was static or not.

src/main/java/io/confluent/idesidecar/restapi/messageviewer/strategy/ConfluentCloudConsumeStrategy.java
rohitsanj: Makes sense. Changed log level to debug and added more info to message.

src/main/java/io/confluent/idesidecar/restapi/messageviewer/strategy/ConfluentCloudConsumeStrategy.java
```diff
@@ -222,6 +225,37 @@ private PartitionConsumeData processPartition(
...
+                BASE64_DECODER.decode(header.value()), StandardCharsets.UTF_8
+            );
+          } catch (IllegalArgumentException e) {
+            // For whatever reason, we couldn't decode the Base64 encoded header value.
+            // Perhaps the API changed or the data is corrupted. Either way,
+            // we log a warning and return the raw value.
+            Log.debugf(e, "Failed to base64 decode header value '%s' from Confluent Cloud" +
+                    "(partition: %d, offset: %d)",
+                header.value(), record.partitionId(), record.offset()
+            );
```
flippingbits, line 246: uber-nit:

```suggestion
            // we log a debug message and return the raw value.
            Log.debugf(
                e,
                "Failed to base64 decode header value '%s' from Confluent Cloud "
                    + "(partition: %d, offset: %d)",
                header.value(),
                record.partitionId(),
                record.offset()
            );
```

flippingbits (approved):
Thanks for the update, @rohitsanj! LGTM.

src/main/java/io/confluent/idesidecar/restapi/messageviewer/strategy/ConfluentCloudConsumeStrategy.java
rohitsanj: Ha, all good. I've addressed it. Thanks!

Issue #181, opened 2024-11-22
Check connection statuses asynchronously


flippingbits:
## Status Quo

At the moment, the endpoints `GET /gateway/v1/connections` and `GET /gateway/v1/connections/:id` (implicitly) call [DirectConnectionState#getConnectionStatus()](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java#L117), which synchronously checks the status of the connection. This can become problematic and lead to high request execution times, at least for the endpoint `GET /gateway/v1/connections` when users manage multiple direct connections.

## Proposal

We should migrate to asynchronously checking the statuses of connections in a background thread, similar to how we [refresh tokens of Confluent Cloud connections](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/auth/RefreshCCloudTokensBean.java#L39). This would allow the endpoints `GET /gateway/v1/connections` and `GET /gateway/v1/connections/:id` to _instantly_ return the most recent status of each connection without having to interact with other/remote systems.

We should make the interval used for checking the connection statuses configurable. It probably makes sense to have one configuration property for each connection type. Initially, we can set the different intervals to the same value, e.g., 5 seconds.

The `status` field of Connections should expose when the status has been determined.

rhauch:
#198 should make it easier to automatically keep the connections' status up-to-date.

flippingbits:
Implemented by https://github.com/confluentinc/ide-sidecar/pull/199

PR #198, opened 2024-11-26
GraphQL queries should not block for direct connections that are not connected

### LINKED ISSUE #197: GraphQL queries blocked when fetching resources for direct connections that are not connected
See https://github.com/confluentinc/vscode/issues/685 for more details.

To reproduce:
1. Start local resources
2. Create a direct connection using local Kafka bootstrap server(s) and/or Schema Registry URI
3. Stop local resources
4. Query GraphQL resources again, which blocks

Logs include
```
WARN  [io.ver.cor.imp.BlockedThreadChecker] (vertx-blocked-thread-checker) Thread Thread[vert.x-eventloop-thread-3,5,main] has been blocked for 58351 ms, time limit is 2000 ms: io.vertx.core.VertxException: Thread blocked
```

rhauch:
Resolves #197. Also relates to #181

## Summary of Changes

Modifies `ConnectionState` as follows:
* add a new `cachedStatus` field to store the most recent status
* renames the asynchronous `getConnectionStatus(): Future<ConnectionStatus>` method to `refreshStatus(): Future<ConnectionStatus>`, changed the implementation to:
  * delegate to a new protected method called `doRefreshStatus()` that subclasses override
  * update the `cachedStatus` field
* add a new *synchronous* `getStatus(): ConnectionStatus` method that returns the `cachedStatus` value

All the uses of the old async `getConnectionStatus(): Future<ConnectionStatus>` have been changed to use the newer async `refreshConnectionStatus(): Future<ConnectionStatus>` method.

The `DirectConnectionState` class was changed to have two convenience methods:
* `isKafkaConnected()` returns true when `status.kafkaCluster().isConnected()` is true; and
* `isSchemaRegistryConnected()` returns true when `status.kafkaCluster().isConnected()` is true

When we want to implement the async status update mechanism (#181), that mechanism only has to call the async `refreshConnectionStatus()`, which will then automatically cache the status in the `ConnectionState` instance.

This commit also changes `RealDirectFetcher` to use the two convenience methods to check the status of the connection’s Kafka cluster and SR *before* using the AdminClient or SR client, respectively, to construct the corresponding cluster information. Only when the status is “SUCCESS” will the clients be used; in all other cases, the `RealDirectFetcher` quickly returning no cluster information.

Added new unit tests for `RealDirectFetcher` to verify the behavior for various statuses and error conditions.

This reduces the risk of blocked threads when the connection's status is up-to-date and known to be good or bad. Right now any `GET /gateway/v1/connections` request will cause all connections' status to be updated, and `GET /gateway/v1/connections/{id}` request will cause that connection's status to be updated.

It still is relatively easy to get out of sync. For example, blocked queries are still a risk if the extension issues GraphQL queries some time after either of the two `GET` requests are issued, or when the connectivity to the external Kafka and/or SR clusters are affected (e.g., by stopping the external Kafka and/or SR clusters) without refreshing the statuses.

We really need #181 to help the sidecar maintain an up-to-date status for all connections.

## Pull request checklist

- Tests:
    - [x] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/connections/CCloudConnectionState.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/connections/CCloudConnectionState.java
@@ -75,7 +75,7 @@ public class CCloudConnectionState extends ConnectionState {
    * @return status of connection
    */
   @Override
-  protected Future<ConnectionStatus> doRefreshConnectionStatus() {
+  protected Future<ConnectionStatus> doRefreshStatus() {
     var missingTokens = new HashSet<String>();
     if (oauthContext.getRefreshToken() == null) {
       missingTokens.add("Refresh token");
--- a/src/main/java/io/confluent/idesidecar/restapi/connections/CCloudConnectionState.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/connections/CCloudConnectionState.java
@@ -46,7 +46,7 @@ public class CCloudConnectionState extends ConnectionState {
   }
 
   @Override
-  public ConnectionStatus getInitialStatus() {
+  protected ConnectionStatus getInitialStatus() {
     return INITIAL_STATUS;
   }
 
@@ -75,7 +75,7 @@ public class CCloudConnectionState extends ConnectionState {
    * @return status of connection
    */
   @Override
-  public Future<ConnectionStatus> getConnectionStatus() {
+  protected Future<ConnectionStatus> doRefreshConnectionStatus() {
     var missingTokens = new HashSet<String>();
     if (oauthContext.getRefreshToken() == null) {
       missingTokens.add("Refresh token");
--- a/src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionState.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionState.java
@@ -89,10 +89,10 @@ public abstract class ConnectionState {
   }
 
   /**
-   * Attempt to {@link #doRefreshConnectionStatus() refresh the connection status and
+   * Attempt to {@link #doRefreshStatus() refresh the connection status and
    * update the {@link #getStatus() cached results}.
    *
-   * <p>This method always calls {@link #doRefreshConnectionStatus()} and then on success updates
+   * <p>This method always calls {@link #doRefreshStatus()} and then on success updates
    * the {@link #getStatus() cached connection status}.
    *
    * @return the future that will complete with the updated connection status
@@ -101,7 +101,7 @@ public abstract class ConnectionState {
    */
   public final Future<ConnectionStatus> refreshStatus() {
     // Always set the cached status when the future completes successfully
-    return doRefreshConnectionStatus().onSuccess(this.cachedStatus::set);
+    return doRefreshStatus().onSuccess(this.cachedStatus::set);
   }
 
   /**
@@ -113,7 +113,7 @@ public abstract class ConnectionState {
    * @see #refreshStatus()
    * @see #getStatus()
    */
-  protected Future<ConnectionStatus> doRefreshConnectionStatus() {
+  protected Future<ConnectionStatus> doRefreshStatus() {
     return Future.succeededFuture(getInitialStatus());
   }
 
--- a/src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionState.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionState.java
@@ -74,7 +74,7 @@ public abstract class ConnectionState {
    *
    * @return the connection status; never null
    */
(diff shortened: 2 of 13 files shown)


src/main/java/io/confluent/idesidecar/restapi/connections/CCloudConnectionState.java
```diff
@@ -46,7 +46,7 @@ public class CCloudConnectionState extends ConnectionState {
   }
 
   @Override
-  public ConnectionStatus getInitialStatus() {
+  protected ConnectionStatus getInitialStatus() {
```
rhauch, line 49: This is only used at instantiation time and may need to be overridden by subclasses, thus `protected` is more constrained.

src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionState.java
```diff
@@ -65,14 +69,54 @@ public void setSpec(ConnectionSpec in) {
...
+   * implement the actual connection status refresh logic.
+   *
+   * @return the future that will complete with the updated connection status
+   * @see #refreshConnectionStatus()
+   * @see #getConnectionStatus()
+   */
+  protected Future<ConnectionStatus> doRefreshConnectionStatus() {
+    return Future.succeededFuture(getInitialStatus());
+  }
+
```
rhauch, line 119: Because we don't want subclasses being responsible for updating the status, I chose to make the public `refreshConnectionStatus()` method _final_, and to move the overridable logic to the new protected `doRefreshConnectionStatus()` method.

I'm not sure about method names, and I was thinking that maybe `doRefreshConnectionStatus()` should be made abstract. WDYT?

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -72,14 +72,34 @@ public DirectConnectionState(
...
+  /**
+   * Check if the connection to the Schema Registry is currently connected.
+   *
+   * @return true if the Kafka component is connected
+   */
+  public boolean isSchemaRegistryConnected() {
+    var status = getConnectionStatus();
+    return status.schemaRegistry() != null && status.schemaRegistry().isConnected();
+  }
+
```
rhauch, line 102: Convenience methods used at the moment in `RealDirectFetcher`, but adding these also makes it easier to override behavior in the `RealDirectFetcherTest` unit tests.

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealDirectFetcher.java
```diff
@@ -164,6 +169,11 @@ protected Uni<DirectKafkaCluster> getKafkaCluster(
   public Uni<DirectSchemaRegistry> getSchemaRegistry(String connectionId) {
     var state = connections.getConnectionState(connectionId);
     if (state instanceof DirectConnectionState directState) {
+      if (!directState.isSchemaRegistryConnected()) {
+        // Either there is no Kafka cluster configured or it is not connected, so return no info
```
flippingbits, line 173:
```suggestion
        // Either there is no Schema Registry configured or it is not connected, so return no info
```

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealDirectFetcher.java
```diff
@@ -112,6 +112,11 @@ public List<DirectConnection> getConnections() {
   public Uni<DirectKafkaCluster> getKafkaCluster(String connectionId) {
     var state = connections.getConnectionState(connectionId);
     if (state instanceof DirectConnectionState directState) {
+      if (!directState.isKafkaConnected()) {
+        // Either there is no Kafka cluster configured or it is not connected, so return no info
+        Log.debugf("Skipping connection '%s' since Kafka is not connected.", connectionId);
+        return Uni.createFrom().nullItem();
+      }
```
rhauch, line 119: We might also consider either doing the same logic in other `Real*Fetcher` implementations, or moving this logic into the different GraphQL resources. (Seems better here, but I'm open to other suggestions.)

flippingbits (approved):
LGTM, @rhauch!

As discussed offline, there is quite some overlap between this PR and #199. We can merge your PR first and I'll then update #199 so that it's aligned with your changes.

rhauch:
Thanks for the review, @flippingbits. I fixed the comment error you caught, and I shortened some of the status-related methods that this PR already changed, taking inspiration from your shorter names in #199. Please take another look when you get a chance. Since you already approved it and these more recent changes are minor, I'm gonna mark this PR for merging as soon as it's approved.

flippingbits (approved):
Thanks for the update, @rhauch. LGTM!

PR #199, opened 2024-11-27
DTX-181: Check connection statuses asynchronously

flippingbits:
## Summary of Changes

This PR migrates the ide-sidecar from synchronously to asynchronously checking the statuses of connections, similar to how we [refresh tokens of Confluent Cloud connections](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/auth/RefreshCCloudTokensBean.java#L39). This allows the endpoints `GET /gateway/v1/connections` and `GET /gateway/v1/connections/:id` to _instantly_ return the most recent **cached** status of each connection without having to interact with remote systems or blocking the processing of the request.

The interval used for refreshing the connection statuses are configurable via the config properties:

* `ide-sidecar.connections.ccloud.refresh-status-interval-seconds`
* `ide-sidecar.connections.confluent-local.refresh-status-interval-seconds`
* `ide-sidecar.connections.direct.refresh-status-interval-seconds`

By default, these config properties are set to 5 seconds. **Setting the intervals to a low value** will allow the endpoints `GET /gateway/v1/connections` and `GET /gateway/v1/connections/:id` to report _fresh_ connection statuses but result in a lot of status checks, potentially putting quite some load on the external Kafka clusters and Schema Registries. **Setting the intervals to a high value** will put less load on external systems but result in the endpoints often reporting outdated connection statuses.

Before this change, users could call the endpoints `GET /gateway/v1/connections` or `GET /gateway/v1/connections/:id` to synchronously check the status of a connection. After this change, this is no longer possible. If they really need to synchronously check the status of a connection (config), they could call the endpoint `POST /gateway/v1/connections` with the query param `?dryRun=true`. Note that this does not persist the connection and does not allow to synchronously check the status of a persisted connection. It's probably only useful for validating the configuration of a _Direct_ connection. See #184 for more details.

With this change, writing tests got more complicated. We can no longer create or update a connection and instantly check its status. Often times, we need to wait for the refresh of the connection status before verifying assertions. In unit tests, we can call the connection's `#refreshStatus` method and verify the assertions once the returned `Future` has been completed. In integration tests, we can poll the connection using Awaitility and proceed when the status has been updated.

When working on this PR, I noticed that `@MockInject`ing beans into beans that define `@Scheduled` methods does sometimes but not always work, and thus leads to flaky tests. One existing example is the [RefreshCCloudTokensBeanTest](https://github.com/confluentinc/ide-sidecar/blob/main/src/test/java/io/confluent/idesidecar/restapi/auth/RefreshCCloudTokensBeanTest.java), which `@MockInject`s the `ConnectionStateManager`, another one would be the `RefreshConnectionStatusesTest` added by this PR. I have not yet figured out why this is the case (maybe it's the execution order of the tests?) but have moved to passing manually-created mocks to the tested bean, which is partially bypassing CDI in the respective tests.

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


flippingbits:
Mh, it looks like `RefreshCCloudTokensBeanTest` became flaky. I'll take a closer look.

src/test/java/io/confluent/idesidecar/restapi/auth/RefreshCCloudTokensBeanTest.java
```diff
@@ -28,6 +29,7 @@ class RefreshCCloudTokensBeanTest {
   RefreshCCloudTokensBean refreshCCloudTokensBean;
 
   @Test
+  @RetryingTest(5)
```
flippingbits, line 32: I added the `RetryingTest` annotation because I noticed that this test randomly fails from time to time. We should try to figure out the underlying issue once time permits.

src/test/java/io/confluent/idesidecar/restapi/auth/RefreshCCloudTokensBeanTest.java
```diff
@@ -45,6 +47,7 @@ void refreshTokensShouldConsiderOnlyCCloudConnections() {
 
   @Test
   @DisabledIfSystemProperty(named = "os.name", matches = ".*Windows.*")
+  @RetryingTest(5)
```
flippingbits, line 50: Same as above.

src/test/java/io/confluent/idesidecar/restapi/auth/RefreshCCloudTokensBeanTest.java
```diff
@@ -27,7 +28,7 @@ class RefreshCCloudTokensBeanTest {
   @Inject
   RefreshCCloudTokensBean refreshCCloudTokensBean;
 
-  @Test
+  @RetryingTest(maxAttempts = 5)
```
flippingbits, line 31: I added the `@RetryingTest` annotation because I noticed that this test fails randomly from time to time. We should try to figure out the underlying issue once time permits.

src/test/java/io/confluent/idesidecar/restapi/auth/RefreshCCloudTokensBeanTest.java
```diff
@@ -43,8 +44,8 @@ void refreshTokensShouldConsiderOnlyCCloudConnections() {
     Mockito.verify(confluentCloudConnection, Mockito.atLeastOnce()).getOauthContext();
   }
 
-  @Test
   @DisabledIfSystemProperty(named = "os.name", matches = ".*Windows.*")
+  @RetryingTest(maxAttempts = 5)
```
flippingbits, line 48: Same as above.

src/main/java/io/confluent/idesidecar/restapi/auth/RefreshCCloudTokensBean.java
```diff
@@ -21,8 +21,12 @@ public class RefreshCCloudTokensBean {
   private static final Integer MAX_TOKEN_REFRESH_ATTEMPTS = ConfigProvider.getConfig()
       .getValue("ide-sidecar.connections.ccloud.refresh-token.max-refresh-attempts", Integer.class);
 
+  final ConnectionStateManager connectionStateManager;
+
   @Inject
-  ConnectionStateManager connectionStateManager;
+  RefreshCCloudTokensBean(ConnectionStateManager connectionStateManager) {
+    this.connectionStateManager = connectionStateManager;
+  }
```
flippingbits, line 29: This change is needed so that we can pass a manually-created mock of the `ConnectionStateManager` when testing this bean, which seems to fix some flaky tests (see https://github.com/confluentinc/ide-sidecar/pull/199#issuecomment-2511388377).

src/main/java/io/confluent/idesidecar/restapi/connections/CCloudConnectionState.java
```diff
@@ -125,12 +125,12 @@ protected Future<ConnectionStatus> doRefreshStatus() {
               return ConnectionStatusBuilder
                   .builder()
                   .ccloud(
-                    new CCloudStatus(
-                        ConnectedState.SUCCESS,
-                        oauthContext.getEndOfLifetime(),
-                        user != null ? user.asUserInfo() : null,
-                        errors
-                    )
+                      new CCloudStatus(
```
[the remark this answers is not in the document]
flippingbits: Good question! In general, we use the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html).

This PR updated this file before https://github.com/confluentinc/ide-sidecar/pull/198 was available and got merged into this PR's branch. It seems my IDE auto-formatted the file per the style guide (see [Indent continuation lines at least +4 spaces](https://google.github.io/styleguide/javaguide.html#s4.5.2-line-wrapping-indent)) when I worked on it. I can revert the formatting change; I hope it's not confusing reviewers.

rhauch (commented):
Thanks, @flippingbits. Overall this looks great. One question I did have is this sentence from the PR description:
> Users are advised to use the endpoint `POST /gateway/v1/connections` with the query param `?dryRun=true` if they need to synchronously check the status of a connection. See #184 for more details.

I don't understand this statement, as `dryRun=true` will not really create a connection. How does this query parameter relate to this PR?

flippingbits:
> I don't understand this statement, as dryRun=true will not really create a connection. How does this query parameter relate to this PR?

@rhauch Before this PR, users (mainly, the extension) could synchronously check the status of a connection by calling the `GET /gateway/v1/connections` endpoint or the `GET /gateway/v1/connections/:id` endpoint. This is no longer possible. If they really need to perform a synchronous status check, they'd have to call the endpoint `POST /gateway/v1/connections` with the query param `?dryRun=true`. I agreed that this is not the same as checking the status of a persisted connection and is probably very confusing. I'll update the statement.

rhauch (approved):
LGTM. Thanks for improving the PR description re my previous comment.

src/test/java/io/confluent/idesidecar/restapi/util/SidecarClient.java
```diff
@@ -304,28 +304,36 @@ public Connection testConnection(ConnectionSpec spec) {
...
+    // connection to the Kafka cluster or Schema registry has been established
+    if (spec.kafkaClusterConfig() != null || spec.schemaRegistryConfig() != null) {
+      await().atMost(Duration.ofSeconds(10)).until(() -> {
+        var connection = given()
+            .when()
+            .get("%s/gateway/v1/connections/%s".formatted(sidecarHost, spec.id()))
+            .then()
+            .statusCode(200)
+            .extract().response().body().as(Connection.class);
+        return connection.status().isConnected();
```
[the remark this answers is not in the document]
flippingbits: Good point! I was aware of the "or" but assumed it's fine to use it as `DirectConnectionState`s `doRefreshStatus()` always checks the status of the Kafka cluster and the SR at the same time. It's probably not a good idea to bake this assumption into the tests. I've updated the method to check the statuses of both the Kafka cluster **and** the Schema Registry.

PR #200, opened 2024-12-02
Cerchie/follow up schema errs class

Cerchie:
## Summary of Changes

@rohitsanj  and @flippingbits, this PR includes suggested changes as a follow-on to https://github.com/confluentinc/ide-sidecar/pull/178

It also includes changes to DirectConnectionState to reflect an apparent refactor in the `/models` folder. 

## Pull request checklist

- Tests:
    - [ ] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
@@ -12,47 +12,59 @@ import java.util.concurrent.ConcurrentHashMap;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 
 /**
- * The `SchemaErrors` class manages schema error caching for different connections.
+ * The SchemaErrors class manages schema error caching for different connections.
  * It uses Caffeine cache to store and retrieve schema errors based on connection and schema IDs.
- *
  * <p>This class is application-scoped and observes connection lifecycle events to manage the cache.
  * It provides methods to read, write, and clear schema errors for specific connections.
+ *
+ * <p>Configuration properties:
+ * <ul>
+ *   <li>`ide-sidecar.schema-fetch-error-ttl`: The time-to-live duration for schema fetch errors in the cache.</li>
+ * </ul>
+ *
+ * <p>Records:
+ * <ul>
+ *   <li>`ConnectionId`: Represents a unique connection identifier.</li>
+ *   <li>`SchemaId`: Represents a unique schema identifier within a cluster.</li>
+ *   <li>`Error`: Represents an error message associated with a schema.</li>
+ * </ul>
+ *
+ * <p>Methods:
+ * <ul>
+ *   <li>`getSubCache(ConnectionId key)`: Retrieves or creates a sub-cache for a specific connection.</li>
+ *   <li>`readSchemaIdByConnectionId(ConnectionId connectionId, SchemaId schemaId)`: Reads a schema error from the cache.</li>
+ *   <li>`writeSchemaIdByConnectionId(ConnectionId connectionId, SchemaId schemaId, Error error)`: Writes a schema error to the cache.</li>
+ *   <li>`clearByConnectionId(ConnectionId connectionId)`: Clears the cache for a specific connection.</li>
+ *   <li>`onConnectionChange(ConnectionState connection)`: Removes the cache by connections id in response to lifecycle events..</li>
+ * </ul>
  */
 
 @ApplicationScoped
 public class SchemaErrors {
 
-  // `ConnectionId` represents a unique connection identifier.
   public record ConnectionId(String id) {
 
   }
 
-  // `SchemaId` represents a unique schema identifier within a cluster.
   public record SchemaId(String clusterId, int schemaId) {
 
   }
 
-  //`Error` represents an error message associated with a schema.
   public record Error(String message) {
 
   }
-  @ConfigProperty(name = "ide-sidecar.api.host")
-  String sidecarHost;
 
-  // The time-to-live duration for schema fetch errors in the cache.
   @ConfigProperty(name = "ide-sidecar.schema-fetch-error-ttl")
   long schemaFetchErrorTtl;
 
   public static final Map<ConnectionId, Cache<SchemaId, SchemaErrors.Error>> cacheOfCaches = new ConcurrentHashMap<>();
 
-  // Retrieves or creates a sub-cache for a specific connection.
   public Cache<SchemaId, Error> getSubCache(ConnectionId key) {
     return cacheOfCaches.computeIfAbsent(
         key,
         k -> Caffeine.newBuilder().expireAfterAccess(Duration.ofSeconds(schemaFetchErrorTtl)).build());
   }
 
-  // Reads a schema error from the cache using both the connection id and the schema id.
   public Error readSchemaIdByConnectionId(
       ConnectionId connectionId,
       SchemaId schemaId
@@ -60,7 +72,6 @@ public class SchemaErrors {
     return getSubCache(connectionId).getIfPresent(schemaId);
   }
 
-  // Writes a schema error to the cache using both the connection id and the schema id.
   public void writeSchemaIdByConnectionId(
       ConnectionId connectionId,
       SchemaId schemaId,
@@ -69,14 +80,12 @@ public class SchemaErrors {
     getSubCache(connectionId).put(schemaId, error);
   }
 
-  // Clears the cache for a specific connection.
   public void clearByConnectionId(
       ConnectionId connectionId
   ) {
     cacheOfCaches.remove(connectionId);
   }
 
-  // Removes the cache by connections id in response to lifecycle events.
   public void onConnectionChange(@ObservesAsync @Lifecycle.Deleted @Lifecycle.Created @Lifecycle.Updated ConnectionState connection) {
     cacheOfCaches.remove(new ConnectionId(connection.getId()));
   }
--- a/src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
@@ -51,14 +51,25 @@ public class SchemaErrors {
 
   public static final Map<ConnectionId, Cache<SchemaId, SchemaErrors.Error>> cacheOfCaches = new ConcurrentHashMap<>();
 
- // getSubCache(ConnectionId key): Retrieves or creates a sub-cache for a specific connection.
+  /**
+   * Retrieves or creates a sub-cache for a specific connection.
+   *
+   * @param key The connection identifier.
+   * @return The sub-cache for the connection.
+   */
   public Cache<SchemaId, Error> getSubCache(ConnectionId key) {
     return cacheOfCaches.computeIfAbsent(
         key,
         k -> Caffeine.newBuilder().expireAfterAccess(Duration.ofSeconds(schemaFetchErrorTtl)).build());
   }
 
-  // readSchemaIdByConnectionId(ConnectionId connectionId, SchemaId schemaId): Reads a schema error from the cache.
+  /**
+   * Reads a schema error from the cache.
+   *
+   * @param connectionId The connection identifier.
+   * @param schemaId     The schema identifier.
+   * @return The schema error, or null if not found.
+   */
   public Error readSchemaIdByConnectionId(
       ConnectionId connectionId,
       SchemaId schemaId
@@ -66,7 +77,13 @@ public class SchemaErrors {
     return getSubCache(connectionId).getIfPresent(schemaId);
   }
 
-  // writeSchemaIdByConnectionId(ConnectionId connectionId, SchemaId schemaId, Error error)`: Writes a schema error to the cache.
+  /**
+   * Writes a schema error to the cache.
+   *
+   * @param connectionId The connection identifier.
+   * @param schemaId     The schema identifier.
+   * @param error        The schema error.
+   */
   public void writeSchemaIdByConnectionId(
       ConnectionId connectionId,
       SchemaId schemaId,
@@ -75,14 +92,22 @@ public class SchemaErrors {
     getSubCache(connectionId).put(schemaId, error);
   }
 
-  // clearByConnectionId(ConnectionId connectionId)`: Clears the cache for a specific connection.
+  /**
+   * Clears the cache for a specific connection.
+   *
+   * @param connectionId The connection identifier.
+   */
   public void clearByConnectionId(
       ConnectionId connectionId
   ) {
     cacheOfCaches.remove(connectionId);
   }
 
-  // onConnectionChange(ConnectionState connection)`: Removes the cache by connections id in response to lifecycle events.
+  /**
+   * Removes the cache by connections id in response to lifecycle events.
+   *
+   * @param connection The connection state.
+   */
   public void onConnectionChange(@ObservesAsync @Lifecycle.Deleted @Lifecycle.Created @Lifecycle.Updated ConnectionState connection) {
     cacheOfCaches.remove(new ConnectionId(connection.getId()));
   }
--- a/src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
@@ -22,32 +22,34 @@ import org.eclipse.microprofile.config.inject.ConfigProperty;
  * <ul>
  *   <li>`ide-sidecar.schema-fetch-error-ttl`: The time-to-live duration for schema fetch errors in the cache.</li>
  * </ul>
- *
- * <p>Records:
- * <ul>
- *   <li>`ConnectionId`: Represents a unique connection identifier.</li>
- *   <li>`SchemaId`: Represents a unique schema identifier within a cluster.</li>
- *   <li>`Error`: Represents an error message associated with a schema.</li>
- * </ul>
  */
 
 @ApplicationScoped
 public class SchemaErrors {
 
+  /**
+   * Represents a unique connection identifier.
+   */
   public record ConnectionId(String id) {
 
   }
 
+  /**
+   * Represents a unique schema identifier within a cluster.
+   */
   public record SchemaId(String clusterId, int schemaId) {
 
   }
 
+  /**
+   * Represents an error message associated with a schema.
+   */
   public record Error(String message) {
 
   }
 
   @ConfigProperty(name = "ide-sidecar.schema-fetch-error-ttl")
-  long schemaFetchErrorTtl;
+  static final Duration schemaFetchErrorTtl;
 
   public static final Map<ConnectionId, Cache<SchemaId, SchemaErrors.Error>> cacheOfCaches = new ConcurrentHashMap<>();
 
--- a/src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
@@ -14,6 +14,7 @@ import org.eclipse.microprofile.config.inject.ConfigProperty;
 /**
  * The SchemaErrors class manages schema error caching for different connections.
  * It uses Caffeine cache to store and retrieve schema errors based on connection and schema IDs.
+ *
  * <p>This class is application-scoped and observes connection lifecycle events to manage the cache.
  * It provides methods to read, write, and clear schema errors for specific connections.
  *
@@ -27,15 +28,6 @@ import org.eclipse.microprofile.config.inject.ConfigProperty;
  *   <li>`ConnectionId`: Represents a unique connection identifier.</li>
  *   <li>`SchemaId`: Represents a unique schema identifier within a cluster.</li>
  *   <li>`Error`: Represents an error message associated with a schema.</li>
- * </ul>
- *
- * <p>Methods:
- * <ul>
- *   <li>`getSubCache(ConnectionId key)`: Retrieves or creates a sub-cache for a specific connection.</li>
- *   <li>`readSchemaIdByConnectionId(ConnectionId connectionId, SchemaId schemaId)`: Reads a schema error from the cache.</li>
- *   <li>`writeSchemaIdByConnectionId(ConnectionId connectionId, SchemaId schemaId, Error error)`: Writes a schema error to the cache.</li>
- *   <li>`clearByConnectionId(ConnectionId connectionId)`: Clears the cache for a specific connection.</li>
- *   <li>`onConnectionChange(ConnectionState connection)`: Removes the cache by connections id in response to lifecycle events..</li>
  * </ul>
  */
 
@@ -59,12 +51,14 @@ public class SchemaErrors {
 
   public static final Map<ConnectionId, Cache<SchemaId, SchemaErrors.Error>> cacheOfCaches = new ConcurrentHashMap<>();
 
+ // getSubCache(ConnectionId key): Retrieves or creates a sub-cache for a specific connection.
   public Cache<SchemaId, Error> getSubCache(ConnectionId key) {
     return cacheOfCaches.computeIfAbsent(
         key,
         k -> Caffeine.newBuilder().expireAfterAccess(Duration.ofSeconds(schemaFetchErrorTtl)).build());
   }
 
+  // readSchemaIdByConnectionId(ConnectionId connectionId, SchemaId schemaId): Reads a schema error from the cache.
   public Error readSchemaIdByConnectionId(
       ConnectionId connectionId,
       SchemaId schemaId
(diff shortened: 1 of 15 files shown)


src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -9,39 +9,58 @@
...
+ * It uses Caffeine cache to store and retrieve schema errors based on connection and schema IDs.
+ *
+ * <p>This class is application-scoped and observes connection lifecycle events to manage the cache.
+ * It provides methods to read, write, and clear schema errors for specific connections.
+ */
 
 @ApplicationScoped
 public class SchemaErrors {
 
+  // `ConnectionId` represents a unique connection identifier.
```
rhauch, line 25: These and most of the other line comments added in this file are probably better captured as JavaDoc comments.

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -9,39 +9,58 @@
...
   public record SchemaId(String clusterId, int schemaId) {
 
   }
 
+  //`Error` represents an error message associated with a schema.
   public record Error(String message) {
 
   }
+  @ConfigProperty(name = "ide-sidecar.api.host")
+  String sidecarHost;
```
rhauch, line 40: How is this used?

src/main/java/io/confluent/idesidecar/restapi/messageviewer/MessageViewerClusterInfoProcessor.java
```diff
@@ -79,25 +79,4 @@ public Future<MessageViewerContext> process(MessageViewerContext context) {
...
-
-  /**
-   * Respond to the connection being disconnected by clearing cached information.
-   *
-   * @param connection the connection that was disconnected
-   */
-  void onConnectionDisconnected(
-      @ObservesAsync @Lifecycle.Disconnected ConnectionState connection
-  ) {
-    schemaErrors.clearByConnectionId(new SchemaErrors.ConnectionId(connection.getId()));
```
rhauch, line 91: The `schemaErrors` field is no longer used and can be removed.

src/main/java/io/confluent/idesidecar/restapi/proxy/ProxyContext.java
```diff
@@ -25,7 +25,7 @@ public class ProxyContext {
   final Buffer requestBody;
   final Map<String, String> requestPathParams;
 
-  final @Nullable String connectionId;
+  final String connectionId;
```
rhauch, line 28: Why this change? Are we changing the contract, or is this a correction (e.g., the field is not really allowed to be null)?

src/test/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializerTest.java
```diff
@@ -363,7 +363,7 @@ record TestCase(
     ) {
       static TestCase nonRetryable(int statusCode) {
         // Expect only 1 try when the status code is non-retryable
-        return new TestCase(statusCode, 1, 1, null);
+        return new TestCase(statusCode, 3, 1, null);
```
rhauch, line 366: I'm curious -- the second parameter is `maxRetries`. Why are we changing this from 1 to 3, when (per the comment) we "expect only 1 try"?

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -9,39 +9,58 @@
 import java.time.Duration;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+/**
+ * The `SchemaErrors` class manages schema error caching for different connections.
```
rhauch, line 15: Backticks don't have any meaning in JavaDoc. Normal convention is to just type the class name as normal text content (suggested), or a link such as `{@link SchemaErrors}` if you *really* want it to be aligned with the type (not really recommended here since it is the JavaDoc for that type).
```suggestion
 * The SchemaErrors class manages schema error caching for different connections.
```

rhauch (commented):
Thanks, @Cerchie. Overall looks good, but I do have a number of mostly minor suggestions.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -199,8 +199,8 @@ protected Future<KafkaClusterStatus> getKafkaConnectionStatus() {
         }
     ).orElseGet(
         () -> {
-            // There is no Kafka cluster configuration, so return a null Kafka status
-            return Future.succeededFuture(null);
+          // There is no Kafka cluster configuration, so return a null Kafka status
+          return Future.succeededFuture(null);
```
rohitsanj, line 203: This diff is extraneous and may be reverted.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -356,4 +356,4 @@ public static Throwable unwrap(Throwable t) {
     }
     return t;
   }
-}
+}
```
rohitsanj, line 359: This diff is extraneous and may be reverted. (A new line must be added back to the end of the file)

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -9,39 +9,58 @@
...
+ * It uses Caffeine cache to store and retrieve schema errors based on connection and schema IDs.
+ *
+ * <p>This class is application-scoped and observes connection lifecycle events to manage the cache.
+ * It provides methods to read, write, and clear schema errors for specific connections.
+ */
```
rohitsanj, line 21: nit: The Javadoc and the class definition should not be separated by whitespace.

rohitsanj (changes requested):
Hi @Cerchie, apart from my offline comment regarding the `DirectConnectionState` refactor which you've already addressed, I have a couple of very minor suggestions. Thanks!

src/main/java/io/confluent/idesidecar/restapi/proxy/ProxyContext.java
Cerchie: This was something I'd added in, probably to make a test pass, but no longer needed, so I reverted it to the prior state

src/test/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializerTest.java
Cerchie: Again, this was a change that slipped in, probably while working on testing: https://github.com/confluentinc/ide-sidecar/pull/178/files#r1861315837

Cerchie: PS-- the second param is `expectedTries`.

Cerchie:
thank you @rohitsanj  and @rhauch for the review! pushed some cleanup to address the comments

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -9,6 +9,35 @@
...
+ * </ul>
+ *
+ * <p>Methods:
+ * <ul>
+ *   <li>`getSubCache(ConnectionId key)`: Retrieves or creates a sub-cache for a specific connection.</li>
+ *   <li>`readSchemaIdByConnectionId(ConnectionId connectionId, SchemaId schemaId)`: Reads a schema error from the cache.</li>
+ *   <li>`writeSchemaIdByConnectionId(ConnectionId connectionId, SchemaId schemaId, Error error)`: Writes a schema error to the cache.</li>
+ *   <li>`clearByConnectionId(ConnectionId connectionId)`: Clears the cache for a specific connection.</li>
+ *   <li>`onConnectionChange(ConnectionState connection)`: Removes the cache by connections id in response to lifecycle events..</li>
+ * </ul>
```
rhauch, line 39: nit: It's more idiomatic to just put these as short JavaDoc comments on each nested type and (public) method, rather than here in the JavaDoc for the class.

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -9,6 +9,35 @@
 import java.time.Duration;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+/**
+ * The SchemaErrors class manages schema error caching for different connections.
+ * It uses Caffeine cache to store and retrieve schema errors based on connection and schema IDs.
+ * <p>This class is application-scoped and observes connection lifecycle events to manage the cache.
```
rhauch, line 17: The linter should complain about `<p>` not being preceded by a blank line, so this is correct:
```suggestion
 *
 * <p>This class is application-scoped and observes connection lifecycle events to manage the cache.
```

src/test/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializerTest.java
rhauch: Hmm, I think the second parameter is `maxRetries` per [these lines](https://github.com/confluentinc/ide-sidecar/blob/7574242da2435a9aab6e68e1ac91a5be41473f42/src/test/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializerTest.java#L358-L363) just above this change:
```
    record TestCase(
        int statusCode,
        int maxRetries,
        int expectedTries,
        Boolean isKey
    ) {
      ...
```

rhauch (commented):
Thanks for the quick updates, @Cerchie. A few more suggestions below.

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -25,23 +46,27 @@ public record Error(String message) {
...
-  public static Cache<SchemaId, Error> getSubCache(ConnectionId key) {
+ // getSubCache(ConnectionId key): Retrieves or creates a sub-cache for a specific connection.
+  public Cache<SchemaId, Error> getSubCache(ConnectionId key) {
     return cacheOfCaches.computeIfAbsent(
         key,
-        k -> Caffeine.newBuilder().expireAfterAccess(SCHEMA_FETCH_ERROR_TTL).build());
+        k -> Caffeine.newBuilder().expireAfterAccess(Duration.ofSeconds(schemaFetchErrorTtl)).build());
   }
 
+  // readSchemaIdByConnectionId(ConnectionId connectionId, SchemaId schemaId): Reads a schema error from the cache.
```
rohitsanj, line 61: We typically try to use [Javadocs](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html) style comments to document functionality. The comment used here `//` is a [single-line (or end-of-line) comment](https://www.oracle.com/java/technologies/javase/codeconventions-comments.html) and we reserve those for clarifying explanations embedded in code implementations.

```
/**
* Reads a schema error from the cache.
*/
```

Cerchie: thanks! pushed a change

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -9,6 +9,27 @@
...
+ * <ul>
+ *   <li>`ide-sidecar.schema-fetch-error-ttl`: The time-to-live duration for schema fetch errors in the cache.</li>
+ * </ul>
+ *
+ * <p>Records:
+ * <ul>
+ *   <li>`ConnectionId`: Represents a unique connection identifier.</li>
+ *   <li>`SchemaId`: Represents a unique schema identifier within a cluster.</li>
+ *   <li>`Error`: Represents an error message associated with a schema.</li>
+ * </ul>
```
rhauch, line 31: Can we please move these to JavaDoc on the record types themselves?

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -25,23 +46,44 @@ public record Error(String message) {
 
   }
 
-  private static final Duration SCHEMA_FETCH_ERROR_TTL = Duration.ofSeconds(30);
+  @ConfigProperty(name = "ide-sidecar.schema-fetch-error-ttl")
+  long schemaFetchErrorTtl;
```
rhauch, line 50: This still can be static and final, and we can still have this be a `Duration` type. See other examples in the codebase.

src/main/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializer.java
```diff
@@ -61,7 +61,7 @@ public class RecordDeserializer {
       .asMap("ide-sidecar.serde-configs");
   private final int schemaFetchMaxRetries;
 
-  SchemaErrors schemaErrors;
+  private final SchemaErrors schemaErrors;
```
rhauch, line 64: Nit: static fields should be separated from and before non-final fields.

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -25,23 +46,44 @@ public record Error(String message) {
...
+   * @param connectionId The connection identifier.
+   * @param schemaId     The schema identifier.
+   * @return The schema error, or null if not found.
+   */
   public Error readSchemaIdByConnectionId(
       ConnectionId connectionId,
       SchemaId schemaId
   ) {
     return getSubCache(connectionId).getIfPresent(schemaId);
   }
```
rhauch, line 78: The record types primarily exist to simplify the definition of the map of maps, so we can be don't need to expose them outside the `SchemaErrors` class. Having these methods _require_ the record types makes these methods harder to use. So we could either:
1. Refactor the public methods to not use the record types; or
2. Add additional methods that don't use record types

For example, option 2 would add methods like:
```
  public Error readSchemaIdByConnectionId(String connectionId, String clusterId, String schemaId) {
    return readSchemaIdByConnectionId(
        new ConnectionId(connectionId),
        new SchemaId(clusterId, schemaId)
    );
  }
```

src/main/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializer.java
```diff
@@ -225,15 +225,14 @@ public DecodedResult deserialize(
...
     if (schemaErrors.readSchemaIdByConnectionId(new SchemaErrors.ConnectionId(context.getConnectionId()), schemaId) != null) {
       return new DecodedResult(
           // If the schema fetch failed, we can't decode the data, so we just return the raw bytes.
           // We apply the encoderOnFailure function to the bytes before returning them.
           onFailure(encoderOnFailure, bytes),
-          String.valueOf(
-              schemaErrors.readSchemaIdByConnectionId(new SchemaErrors.ConnectionId(context.getConnectionId()), schemaId))
-      );
+          schemaErrors.readSchemaIdByConnectionId(connectionId, schemaId).message());
     }
```
rhauch, line 236: We really should look this up once and use the result multiple times. If we read multiple times, there's a chance that we'll get different results back.
```
var schemaId = getSchemaIdFromRawBytes(bytes);
var error = schemaErrors.readSchemaIdByConnectionId(
    contex.getConnectionId(),
    context.getClusterId(),
    schemaId
);
if (error != null) {
  return new DecodedResult(
      onFailure(encoderOnFailure, bytes),
      error.message()
  );
}
```

rhauch (commented):
Thanks for the updates, @Cerchie.

rohitsanj (approved):
Thanks for the updates! LGTM.

Let's wait for @rhauch to bless as well.

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
Cerchie: I might be missing something... wouldn't this recurse as written?

rhauch: No, the `readSchemaIdByConnectionId(String, String, String)` method can call the `readSchemaIdByConnectionId(ConnectionId, SchemaId)` method and it will not recurse since they're different signatures.

Cerchie:
> Thanks for the updates, @Cerchie.

Thank you! with the exception of resolving the issue of exposing the records outside the class, I've addressed the comments and reviewed a diff to try to catch style issues before pushing up the latest commits.

src/main/resources/application.yml
```diff
@@ -67,6 +68,7 @@ vscode:
     version: unknown
   version: unknown
 ide-sidecar:
+  schema-fetch-error-ttl: 30
```
rhauch, line 71: Nit: can we put this lower, maybe just before `schema-fetch-retry` (since before it would be in proper sort-order)?

src/main/resources/application.yml
```diff
@@ -2,6 +2,7 @@
 ---
 "%test":
   ide-sidecar:
+    schema-fetch-error-ttl: 30
```
rhauch, line 5: Nit: can we put this lower, maybe just before `schema-fetch-retry` (since before it would be in proper sort-order)?

src/test/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializerTest.java
```diff
@@ -46,7 +46,7 @@ public class RecordDeserializerTest {
 
   private static final SchemaErrors.SchemaId VALID_SCHEMA_ID = new SchemaErrors.SchemaId("fake_cluster_id", 10008);
 
-  SchemaErrors.ConnectionId CONNECTION_1_ID = new SchemaErrors.ConnectionId("c1");
+  String CONNECTION_1_ID = "c1";
```
rhauch, line 49: We're considering this a constant if the name is in all caps, so let's make it final
```suggestion
  static final String CONNECTION_1_ID = "c1";
```

src/main/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializer.java
```diff
@@ -345,7 +350,12 @@ private void cacheSchemaFetchError(
         schemaId.schemaId(),
         e.getMessage()
     ));
-    schemaErrors.writeSchemaIdByConnectionId(new SchemaErrors.ConnectionId(context.getConnectionId()), schemaId, errorMessage);
+    schemaErrors.writeSchemaIdByConnectionId(
+        context.getConnectionId(),
+        String.valueOf(schemaId.schemaId()),
```
rhauch, line 355: I think this was a miss on my part, but we shouldn't really be converting the schema IDs to strings and then back to integers. If we need them as integers, why don't we expect an integer parameter?

src/test/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializerTest.java
```diff
@@ -63,7 +63,7 @@ public class RecordDeserializerTest {
       null,
       null,
       null,
-      CONNECTION_1_ID,
+      new SchemaErrors.ConnectionId(CONNECTION_1_ID),
```
rhauch, line 66: I recently suggested think NOT exposing `SchemaErrors.ConnectionId` or `SchemaErrors.SchemaId` outside of `SchemaErrors`. If that's the case, why do we need (or want) the MessageViewerContext to have an instance of ``SchemaErrors.ConnectionId`?

I think it breaks encapsulation and abstraction if `MessageViewerContext` has a field of type `SchemaErrors.ConnectionId` just to hold onto a connection ID, as the connection ID is not at all related to SchemaErrors.

src/main/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializer.java
```diff
@@ -225,14 +225,19 @@ public DecodedResult deserialize(
       return result.get();
     }
     SchemaErrors.SchemaId schemaId = new SchemaErrors.SchemaId(context.getClusterId(), getSchemaIdFromRawBytes(bytes));
+    SchemaErrors.ConnectionId connectionId = new SchemaErrors.ConnectionId(context.getConnectionId());
```
rhauch, line 228: Again, why does `RecordDeserializer` know anything about `SchemaErrors.SchemaId` or `SchemaErrors.ConnectionId`? Those types should be _internal_ / _private_ to the `SchemaErrors` class.

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
```diff
@@ -9,54 +9,124 @@
...
+  public static final Map<ConnectionId, Cache<SchemaId, Error>> cacheOfCaches = new ConcurrentHashMap<>();
 
-  public static Cache<SchemaId, Error> getSubCache(ConnectionId key) {
+  /**
+   * Retrieves or creates a sub-cache for a specific connection.
+   *
+   * @param key The connection identifier.
+   * @return The sub-cache for the connection.
+   */
+  public Cache<SchemaId, Error> getSubCache(ConnectionId key) {
```
rhauch, line 67: Why is this method public? IIUC, it's not used outside of this class, so we can do this:
```suggestion
  private Cache<SchemaId, Error> getSubCache(ConnectionId key) {
```
If it were used in testing, we could make it _package protected_.

rhauch (commented):
A few more suggestions. The big question is: why are the `ConnectionId` and `SchemaId` record types even exposed outside of `SchemaErrors`. As mentioned before, these types only exist to make the implementation of `SchemaErrors` more clear. 
We only added them because it's very confusing to have `Map<String, Map<String, String>` with the second-level key be a string composed from two other strings. Using record types makes this much clearer: `Map<ConnectionId, Map<SchemaId, Error>`.

And a recent suggestion was to remove the `ConnectionId` and `SchemaId` from the public method signatures, _so that we can make these record types private_. We haven't yet made the record types private, and need to do that.

The `Error` record type can be public, because it valuable outside to the components that use the `SchemaErrors` object.

One might argue that we should have used a record type for connection ID, but if we were to do that we'd want that record type to be defined in a common place and definitely NOT scoped within `SchemaErrors`, because we'd want to use it *everywhere*. But that's a big change, and at this point it's not worth it as we don't need the type safety everywhere in the codebase. (As mentioned above, it's _locally_ valuable in `SchemaErrors` to help disambiguate the types in the nested caches.)

Cerchie:
> A few more suggestions. The big question is: why are the `ConnectionId` and `SchemaId` record types even exposed outside of `SchemaErrors`. As mentioned before, these types only exist to make the implementation of `SchemaErrors` more clear. We only added them because it's very confusing to have `Map<String, Map<String, String>` with the second-level key be a string composed from two other strings. Using record types makes this much clearer: `Map<ConnectionId, Map<SchemaId, Error>`.
> 
> And a recent suggestion was to remove the `ConnectionId` and `SchemaId` from the public method signatures, _so that we can make these record types private_. We haven't yet made the record types private, and need to do that.
> 
> The `Error` record type can be public, because it valuable outside to the components that use the `SchemaErrors` object.
> 
> One might argue that we should have used a record type for connection ID, but if we were to do that we'd want that record type to be defined in a common place and definitely NOT scoped within `SchemaErrors`, because we'd want to use it _everywhere_. But that's a big change, and at this point it's not worth it as we don't need the type safety everywhere in the codebase. (As mentioned above, it's _locally_ valuable in `SchemaErrors` to help disambiguate the types in the nested caches.)

Ah I missed making those records private after adjusting the methods. Thanks for your patience. Moved the ttl setting to a more appropriate place, made the records private, and also made the `schemaId` an int.

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaErrors.java
rhauch: +1

src/test/java/io/confluent/idesidecar/restapi/integration/AbstractIT.java
```diff
@@ -63,15 +63,15 @@
  */
 public abstract class AbstractIT extends SidecarClient implements ITSuite {
 
-  static SchemaErrors.ConnectionId VALID_CONNECTION_ID = new SchemaErrors.ConnectionId("c1");
+  private static String VALID_CONNECTION_ID = "c1";
```
rhauch, line 66:
```suggestion
  private static final String VALID_CONNECTION_ID = "c1";
```

rhauch (commented):
Two super minor nits, but otherwise looks great.

rhauch (approved):
LGTM. Nice work, @Cerchie!

PR #201, opened 2024-12-03
Implement support for TLS config, OAuth credentials and setup CP integration tests

rohitsanj:
## Summary of Changes

- Add cp-demo testcontainers, along with a script to set it up from scratch -- works locally and on CI.
- Added make targets to start/stop cp-demo.
- Added make targets to store/load the Docker images into/from Semaphore cache. (I noticed we weren't caching confluent-local, so I've added that too.)
- Bring @rhauch's implementation of mTLS and OAuth credentials from #211 and tweaked as follows. Mutual TLS is no longer a "credentials" type, but rather a sub-field called `ssl` under `ConnectionSpec.KafkaClusterConfig` as well as `ConnectionSpec.SchemaRegistryConfig`. The `ssl` config may be set along with any credentials type (basic, apikey/secret, oauth) and can even be set in the absence of creds.
- The field `ssl` holds all configs related to TLS. (h/t @flippingbits for this section)
  - It can be used to disable the usage of TLS via the boolean field enabled, in which case the sidecar uses an unencrypted but potentially authenticated channel to interact with the Kafka cluster.
  - It accepts a keystore via the field `keystore`, a truststore via the field `truststore`, and allows to disable hostname verification via the boolean field `verify_hostname`. 
  - If `ssl` is set to null, the sidecar uses an encrypted channel to interact with the Kafka cluster, verifies the hostname, and does not take any custom keystore/truststore into account.

Follow up items:

- Add a new Credentials type called `MutualTLSCredentials` starting with a single field such as `derived` that is always set to `true`. The SSL configuration for mTLS will still need to be passed in the `kafka_cluster.ssl` (or `schema_registry.ssl`) object. -- #237 
- The OAuth credentials type is untested, we can add integration tests for this in a follow up PR. -- #238 
- ~Fix pom.xml configuration to ensure we run integration tests just once, not twice.~ (Fixed in #224)
- Replace ZK with KRaft, start just one Kafka broker -- https://github.com/confluentinc/ide-sidecar/issues/239

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [x] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/test/java/io/confluent/idesidecar/restapi/util/CPDemoTestEnvironment.java
```diff
@@ -0,0 +1,113 @@
...
+import io.confluent.idesidecar.restapi.util.cpdemo.SchemaRegistryContainer;
+import io.confluent.idesidecar.restapi.util.cpdemo.ZookeeperContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.containers.wait.strategy.Wait;
+
+import java.util.Optional;
+
+/**
+ * A {@link TestEnvironment} that starts a CP Demo environment with a two-node Kafka cluster,
+ * Zookeeper, OpenLDAP, and Schema Registry.
```
flippingbits, line 15: For the sake of faster tests and a better memory footprint, what about running a **one**-node Kafka cluster in [KRaft mode](https://docs.confluent.io/platform/current/installation/docker/config-reference.html#required-ak-configurations-for-kraft-mode)?

https://docs.confluent.io/platform/current/installation/docker/config-reference.html#required-ak-configurations-for-zk-mode states that `As of Confluent Platform 7.5, ZooKeeper is deprecated for new deployments.`

src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java
```diff
@@ -49,7 +51,17 @@ public record ConnectionSpec(
                           + "Apache Kafka cluster.")
     @JsonProperty(KAFKA_CLUSTER_CONFIG_FIELD_NAME) KafkaClusterConfig kafkaClusterConfig,
     @Schema(description = "The details for connecting to a Schema Registry.")
-    @JsonProperty(SCHEMA_REGISTRY_CONFIG_FIELD_NAME) SchemaRegistryConfig schemaRegistryConfig
+    @JsonProperty(SCHEMA_REGISTRY_CONFIG_FIELD_NAME) SchemaRegistryConfig schemaRegistryConfig,
```
rhauch, line 54: Ooh, this pattern established above is harder to read. Maybe we should be consistent and have all annotations be on lines before the actual record parameter type and name.
```suggestion
    @JsonProperty(SCHEMA_REGISTRY_CONFIG_FIELD_NAME)
    SchemaRegistryConfig schemaRegistryConfig,
```

src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java
```diff
@@ -49,7 +51,17 @@ public record ConnectionSpec(
...
+    @Schema(
+        description =
+            "Whether to communicate with the cluster over TLS/SSL. Defaults to 'true', "
+                + "but set to 'false' when the cluster does not support TLS/SSL.",
+        defaultValue = KafkaClusterConfig.DEFAULT_SSL_VALUE,
```
rhauch, line 59: The `TLSConfig tlsConfig` parameter is not a boolean, so the `Defaults to ...` and `defaultValue` here are not right.

src/main/java/io/confluent/idesidecar/restapi/credentials/TLSConfig.java
```diff
@@ -0,0 +1,304 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+@Schema(description = "SSL configuration")
+public record TLSConfig(
+    @Schema(description = "Whether to verify the server's SSL certificate. If set to false, "
+        + "the server's SSL certificate will not be verified. Defaults to true.")
```
rhauch, line 26: Can we use the `defaultValue` attribute of the `@Schema` annotation, so the default value is captured in the OpenAPI spec?

src/main/java/io/confluent/idesidecar/restapi/credentials/TLSConfig.java
```diff
@@ -0,0 +1,304 @@
...
+    @Null
+    Boolean verifySslCertificates,
+    @Schema(description =
+        "The trust store configuration for authenticating the server's certificate.")
+    @JsonProperty(value = "truststore")
+    @NotNull
+    TrustStore truststore,
+    @Schema(description = "The key store configuration for "
+        + "authenticating the client to the server.",
+        nullable = true)
```
rhauch, line 37: Should we explain that this need be provided only when using mTLS? Maybe something like:
```suggestion
    @Schema(description = "The key store configuration that will identify and authenticate the client to the server, required for mutual TLS (mTLS)",
        nullable = true)
```

src/main/java/io/confluent/idesidecar/restapi/credentials/TLSConfig.java
```diff
@@ -0,0 +1,304 @@
...
+      )
+      @JsonProperty(value = "password")
+      @Size(max = TRUSTSTORE_PASSWORD_MAX_LEN)
+      @Null
+      Password password,
+
+      @Schema(description = "The file format of the local trust store file", nullable = true)
+      @JsonProperty(value = "type")
+      @Null
+      StoreType type
```
rhauch, line 65: Maybe we should either have a default or require this to be not-null. As discussed earlier, the VS Code extension will likely set this anyway.

src/test/java/io/confluent/idesidecar/restapi/cache/ClientConfiguratorStaticTest.java
```diff
@@ -262,6 +307,114 @@ record TestInput(
...
+//                """.formatted(MTLS_TRUSTSTORE_PASSWORD, MTLS_KEYSTORE_PASSWORD, MTLS_KEY_PASSWORD),
+//            """
+//                schema.registry.url=http://localhost:8081
+//                ssl.truststore.location=/path/to/truststore
+//                ssl.truststore.password=%s
+//                ssl.keystore.location=/path/to/keystore
+//                ssl.keystore.password=%s
+//                ssl.key.password=%s
+//                """.formatted(MTLS_TRUSTSTORE_PASSWORD, MTLS_KEYSTORE_PASSWORD, MTLS_KEY_PASSWORD)
+//        ),
```
rhauch, line 338: This should probably be removed as-is, since we don't have `MutualTLSCredentials` anymore. And we somehow need to change `TestInputs` so that we can specify the TLS config to the `ClientConfigurator` methods.

IIUC, we are never really testing with TLS or with not-verifying certs.

src/main/java/io/confluent/idesidecar/restapi/credentials/Password.java
```diff
@@ -13,6 +14,7 @@
  * returned in an API response.
  */
 @JsonDeserialize(using = Password.Deserializer.class)
+@JsonSerialize(using = Password.Serializer.class)
```
rhauch, line 17: My concern with doing this is that this serializer is used to serialize the `Redactable` objects _in the responses_. We absolutely cannot allow that.

We might need to not register this serializer, and *only register the serializer in the `SidecarClient`'s `ObjectMapper` instance*, something akin to:

```
ObjectMapper mapper = new ObjectMapper();
SimpleModule module = new SimpleModule();
module.addSerializer(Password.class, new Password.Serializer());
module.addSerializer(ApiSecret.class, new ApiSecret.Serializer());
mapper.registerModule(module);
```

src/main/java/io/confluent/idesidecar/restapi/credentials/Redactable.java
```diff
@@ -42,6 +45,19 @@ public T deserialize(JsonParser p, DeserializationContext ctxt) throws IOExcepti
     protected abstract T create(char[] value);
   }
 
+  protected abstract static class BaseSerializer<T extends Redactable> extends JsonSerializer<T> {
```
rhauch, line 48: We should call this `RawSerializer` or something, to make it more obvious this will serialize the **raw value of the `Redactable`**, which is unsafe to use in production.

In fact, we might want to move all this code to the `SidecarClient` class, so that we never have a serializer for `Redactable` subclasses in production code.

src/main/resources/application.yml
```diff
@@ -212,14 +212,15 @@ quarkus:
       # We do not use the Quarkus Kafka devservices container to start the Kafka cluster
       enabled: false
   log:
-    min-level: INFO
+    min-level: DEBUG
```
rhauch, line 215: Let's not include this change.

src/main/resources/application.yml
```diff
@@ -212,14 +212,15 @@ quarkus:
...
     category:
       "org.apache.kafka.clients":
         level: ERROR
       "org.apache.kafka.common":
         level: ERROR
       "io.confluent.kafka.serializers":
         level: ERROR
+    level: DEBUG
```
rhauch, line 223: Let's not include this change.

src/test/java/io/confluent/idesidecar/restapi/cache/ClientConfiguratorStaticTest.java
```diff
@@ -44,15 +48,56 @@ class ClientConfiguratorStaticTest {
...
+          new Password(MTLS_TRUSTSTORE_PASSWORD.toCharArray()),
+          TLSConfig.StoreType.JKS
+      ),
+      new KeyStore(
+          MTLS_KEYSTORE_PATH,
+          new Password(MTLS_KEYSTORE_PASSWORD.toCharArray()),
+          TLSConfig.StoreType.JKS,
+          new Password(MTLS_KEY_PASSWORD.toCharArray())
+      )
+  );
```
rhauch, line 100: We need tests to use this.

src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/ClusterProxyContext.java
```diff
@@ -62,4 +62,5 @@ public String getClusterId() {
   public ClusterType getClusterType() {
     return clusterType;
   }
+
```
rhauch, line 65: Unnecessary

rhauch (commented):
Thanks, @rohitsanj. I took a quick first pass, and wanted to share these questions/suggestions sooner than later.

src/main/java/io/confluent/idesidecar/restapi/proxy/ProxyHttpClient.java
```diff
@@ -3,22 +3,29 @@
...
+    this.vertx = vertx;
   }
 
   public Future<T> send(T context) {
-    return webClientFactory.getWebClient()
-        .requestAbs(context.proxyRequestMethod, context.proxyRequestAbsoluteUrl)
-        .putHeaders(context.proxyRequestHeaders)
-        .sendBuffer(context.proxyRequestBody)
+    var options = webClientFactory.getDefaultWebClientOptions();
+    options.setTrustStoreOptions(context.getTruststoreOptions());
```
rohitsanj, line 23: This needs a `!= null` check

src/main/java/io/confluent/idesidecar/restapi/proxy/ProxyRequestProcessor.java
```diff
@@ -1,22 +1,24 @@
...
 
 /**
  * Generic processor that ships the request to the target server and updates the context with the
  * response.
  *
- * @param <T> The type of the context that must extend {@link ProxyContext}
+ * @param <T> The type of the context that must extend {@link ClusterProxyContext}
  */
-public class ProxyRequestProcessor<T extends ProxyContext> extends
+public class ProxyRequestProcessor<T extends ClusterProxyContext> extends
```
rohitsanj, line 15: This should be reverted to `extends ProxyContext`

src/main/java/io/confluent/idesidecar/restapi/proxy/clusters/processors/ClusterStrategyProcessor.java
```diff
@@ -23,6 +23,7 @@
 @ApplicationScoped
 public class ClusterStrategyProcessor extends
     Processor<ClusterProxyContext, Future<ClusterProxyContext>> {
+
```
rohitsanj, line 26: Remove

src/test/java/io/confluent/idesidecar/restapi/util/SidecarClient.java
```diff
@@ -355,7 +359,7 @@ public Connection createConnection(ConnectionSpec spec) {
     // If the connection spec configures a Kafka cluster or a Schema Registry, wait until the
     // connection to the Kafka cluster or Schema registry has been established
     if (spec.kafkaClusterConfig() != null || spec.schemaRegistryConfig() != null) {
-      await().atMost(Duration.ofSeconds(10)).until(() -> {
+      await().atMost(Duration.ofSeconds(30)).until(() -> {
```
rohitsanj, line 362: Revert this

src/test/java/io/confluent/idesidecar/restapi/util/cpdemo/Constants.java
```diff
@@ -0,0 +1,30 @@
...
+           username="admin"
+           password="admin-secret"
+           user_admin="admin-secret"
+           user_mds="mds-secret";
+      };
+
+      tokenhost.KafkaServer {
+           org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required
+           publicKeyPath="/tmp/conf/public.pem";
+      };
```
rohitsanj, line 27: @rhauch I had to add this for the additional hosts to work. Else, the cp-server container shuts down with "internalhost.KafkaServer not defined in JAAS config"

I don't understand why this wasn't needed for the other existing listeners.

src/main/java/io/confluent/idesidecar/restapi/clients/SchemaRegistryClients.java
```diff
@@ -85,12 +86,21 @@ private SchemaRegistryClient createClient(
...
   ) {
+    var restService = new RestService(srClusterUri);
+    restService.configure(configurationProperties);
+    restService.setHttpHeaders(headers);
+
+    var sslFactory = new SslFactory(configurationProperties);
+    if (sslFactory.sslContext() != null) {
+      restService.setSslSocketFactory(sslFactory.sslContext().getSocketFactory());
+    }
+
```
rohitsanj, line 97: If we don't instantiate the RestService this way and pass it into the `CachedSchemaRegistryClient`, it will fail with `unable to find valid certification path to requested target`. I have no idea why. 

How did I land at this solution? I'd extracted out the code from the constructor to debug what was being set in the `sslFactory` and before I knew it, that had fixed the issue. 🤷 I'll take it.

rohitsanj (commented):
Some notes

src/test/java/io/confluent/idesidecar/restapi/util/cpdemo/Constants.java
rhauch: I'm not really sure why cp-server behaves this way, and I couldn't find anything about `not defined in JAAS config` or `internalhost.KafkaServer`. Maybe the token mechanism specified elsewhere is checking for this in the config? That might explain why it wasn't needed for other listener configs.

src/main/java/io/confluent/idesidecar/restapi/credentials/TLSConfig.java
rohitsanj: > Maybe we should either have a default or require this to be not-null.

How about we do both? The default would be `JKS` but we still require it to be set.

pom.xml
```diff
@@ -3,12 +3,6 @@
   xmlns="http://maven.apache.org/POM/4.0.0"
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <modelVersion>4.0.0</modelVersion>
-  <parent>
-    <groupId>io.confluent</groupId>
-    <artifactId>common-parent</artifactId>
-    <version>7.7.0</version>
-  </parent>
```
flippingbits, line 10: Why did you have to remove the parent POM?

Makefile
```diff
@@ -95,3 +95,93 @@ upload-artifacts-to-github-release:
...
+	./mvnw -s .mvn/settings.xml \
+		-Dexec.mainClass=io.confluent.idesidecar.restapi.util.CPDemoTestEnvironment \
+		-Dexec.classpathScope=test \
+		-Dexec.args=stop \
+		exec:java
+
+
+CONFLUENT_DOCKER_TAG = $(shell yq e '.ide-sidecar.integration-tests.cp-demo.tag' src/main/resources/application.yml | sed 's/^v//')
+# See io.confluent.idesidecar.restapi.util.ConfluentLocalKafkaWithRestProxyContainer
+CONFLUENT_LOCAL_DOCKER_TAG = "7.6.0"
```
rhauch, line 126: [CP 7.8 was released/announced](https://www.confluent.io/blog/introducing-confluent-platform-7-8/) on Tuesday this week. Do we want to stick with 7.6.0 for now and maybe migrate in a subsequent PR?

pom.xml
```diff
@@ -425,6 +419,14 @@
...
             </executions>
           </plugin>
+          <!-- Disable Surefire plugin for native profile -->
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <skip>true</skip>
+            </configuration>
+          </plugin>
```
rhauch, line 429: Nice catch!

src/main/resources/application.yml
```diff
@@ -198,6 +198,10 @@ ide-sidecar:
     max-backoff-ms: 5000
     timeout-ms: 10000
     max-retries: 5
+  integration-tests:
+    # cp-demo configs used by Confluent Platform integration tests
+    cp-demo:
+      tag: v7.7.1
```
rhauch, line 204: Does the this version or the one in the makefile take precedence?

rhauch (commented):
Great work, @rohitsanj! The non-test code changes look good to me, and the test code changes are good, too. I have a few questions below, but nothing major... except two things.

First, by removing the parent POM it is possible (even likely?) that we're now using different versions of dependencies that we've not explicitly set. We should verify that is not the case. 

Second, are we sure we want to do this change with potentially major impact in this PR? What is the impact of postponing this to a subsequent PR? I presume the builds are longer, but by how much compared to current `main`?

I'll be AFK for a few days, so please don't block on me for approval.

flippingbits (commented):
Thanks, @rohitsanj. That's an excellent start. I have a few comments and questions.

At the moment, the test containers are a more or less 1:1 mapping of the cp-demo Docker Compose configuration. WDYT about working on more generic, configurable, and reusable testcontainers for Confluent Platform in the mid/long term, so that we can potentially make them available to other projects?

Makefile
rohitsanj: Yeah, sounds good. We could migrate to 7.8 in a subsequent PR.

src/main/resources/application.yml
rohitsanj: The Makefile parses the version defined here using `yq`.

pom.xml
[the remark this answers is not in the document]
rohitsanj: Answered in PR description:

> It turns out that we were running the integration test suite against the native executable -- not once, but TWICE! Apparently, it was due to the [<id>](https://stackoverflow.com/a/72302045) being set (WTF?) in the failsafe plugin config. I've removed that. **However, that wasn't enough since [confluent/common-parent's pom is also riddled with this oddity](https://github.com/confluentinc/common/blob/v7.7.0/pom.xml#L839). I've removed that as well so ide-sidecar no longer has confluentinc/common-parent as a parent POM.***

pom.xml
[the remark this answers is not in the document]
rohitsanj: Yes, it does disable running unit tests before packaging native executable. Hmm.. perhaps we want to do something like `<skip>${skipSurefire}</skip>` and then only set it to true by adding the argument `-DskipSurefire=true` in the `make mvn-verify-native-executable` but not for `make mvn-package-native`.

src/generated/resources/openapi.json
```diff
@@ -919,6 +919,19 @@
...
+            "type" : "object",
+            "allOf" : [ {
+              "$ref" : "#/components/schemas/TLSConfig"
+            } ],
+            "nullable" : true
+          },
+          "verify_server_certificate_hostname" : {
+            "description" : "Whether to verify the server certificate hostname. Defaults to true if not set.",
+            "default" : true,
+            "type" : "boolean"
```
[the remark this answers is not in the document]
rohitsanj: Okay, we can move it down back into the Kafka/SR config models. No, it doesn't affect any other resource.

src/generated/resources/openapi.json
```diff
@@ -1486,6 +1567,32 @@
...
+            } ]
+          },
+          "keystore" : {
+            "description" : "The key store configuration that will identify and authenticate the client to the server, required for mutual TLS (mTLS)",
+            "type" : "object",
+            "allOf" : [ {
+              "$ref" : "#/components/schemas/KeyStore"
+            } ],
+            "nullable" : true
+          }
```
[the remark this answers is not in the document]
rohitsanj: Hmm, I think this will require some discussion. If we made mTLS a credentials type for Kafka cluster, how do we support the configuration for where a user wants to use Basic/API key-secret/OAuth credentials to authenticate with Kafka over SSL?

> I feel we should, in general, differentiate between mTLS used for authentication and custom certs used for verifying the server's cert. WDYT?

Makes sense, I think. The (simplifying) assumption made was that these will most likely be the same.

src/main/java/io/confluent/idesidecar/restapi/clients/ClientConfigurator.java
```diff
@@ -314,11 +333,34 @@ public static Map<String, Object> getSchemaRegistryClientConfig(
...
+      creds.get().schemaRegistryClientProperties(options).ifPresent(props::putAll);
+      connection.getTLSConfig().ifPresentOrElse(
+          ignored -> props.put("security.protocol", "SASL_SSL"),
+          () -> props.put("security.protocol", "SASL_PLAINTEXT")
+      );
+    } else {
+      connection.getTLSConfig().ifPresent(
+          ignored -> props.put("security.protocol", "SSL")
+      );
+    }
```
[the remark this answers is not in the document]
rohitsanj: Ah, thanks. I will remove this section for SR.

src/test/java/io/confluent/idesidecar/restapi/integration/ConfluentPlatformIT.java
```diff
@@ -11,25 +30,91 @@ public class ConfluentPlatformIT {
...
     TEST_ENVIRONMENT.start();
   }
 
+  @AfterAll
+  static void afterAll() {
+    // Shutdown the test environment after all tests have run
+    // This behavior is only enabled on CI
+    if (System.getenv("CI") != null && System.getenv("CI").equals("true")) {
+      TEST_ENVIRONMENT.shutdown();
+    }
```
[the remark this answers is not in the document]
rohitsanj: They wouldn't automatically shutdown. You'd have to shut them down manually from Docker desktop or use the `make cp-demo-stop`. Like we discussed in a meeting earlier, this behavior can likely be opt-in through a developer managed YAML config file or something.

src/test/java/io/confluent/idesidecar/restapi/util/CPDemoTestEnvironment.java
```diff
@@ -0,0 +1,448 @@
...
+ */
+@SetEnvironmentVariable(key = "TESTCONTAINERS_REUSE_ENABLE", value = "true")
+// We want to manage cleanup ourselves.
+public class CPDemoTestEnvironment implements TestEnvironment {
+  private Network network;
+  private ToolsContainer tools;
+  private ZookeeperContainer zookeeper;
+  private OpenldapContainer ldap;
+  private CPServerContainer kafka1;
+  private CPServerContainer kafka2;
```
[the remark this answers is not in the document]
rohitsanj: Yes, for sure -- we can lay out a plan to make this more configurable/generic like you mention. However, I'd like to follow YAGNI for now and keep this more-or-less hardcoded implementation of cp-demo as is while it serves our need for test coverage over CP features.

rohitsanj:
@rhauch: I [compared](https://gist.github.com/rohitsanj/6211682c3e063f3233f8cdaacffa8267) the output of `make mvn-package-native-sources-only && ls target/native-sources/lib` for with parent POM set to confluentinc/common-parent, and for without:

```
❯ diff yes-parent-deps.txt no-parent-deps.txt
27c27
< com.google.code.gson.gson-2.9.0.jar
---
> com.google.code.gson.gson-2.11.0.jar
30,31c30
< com.google.guava.guava-32.0.1-jre.jar
< com.google.guava.listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar
---
> com.google.guava.guava-33.2.1-jre.jar
33c32
< com.google.protobuf.protobuf-java-3.25.3.jar
---
> com.google.protobuf.protobuf-java-3.25.5.jar
230c229
< org.apache.kafka.kafka-clients-7.7.0-ccs.jar
---
> org.apache.kafka.kafka-clients-3.7.1.jar
264,265c263,264
< org.scala-lang.scala-library-2.13.10.jar
< org.slf4j.slf4j-api-1.7.36.jar
---
> org.scala-lang.scala-library-2.13.14.jar
> org.slf4j.slf4j-api-2.0.6.jar
268c267
< org.yaml.snakeyaml-2.0.jar
---
> org.yaml.snakeyaml-2.2.jar
```

As you rightly suspected, there are a few dependency changes. The list of changes indicate how to go from yes-parent-deps.txt to no-parent-deps.txt, with `<` meaning line removed, `>` meaning line added.

One difference to highlight is the `org.apache.kafka:kafka-clients` package -- with the parent POM it was Confluent's [7.7.0-ccs version](https://mvnrepository.com/artifact/org.apache.kafka/kafka-clients/7.7.0-ccs), and without it we seem to pull Apache's [3.7.1 version](https://mvnrepository.com/artifact/org.apache.kafka/kafka-clients/3.7.1).

> Second, are we sure we want to do this change with potentially major impact in this PR? What is the impact of postponing this to a subsequent PR? I presume the builds are longer, but by how much compared to current main?

Given the analysis above and the potential unknown impact beyond just a few dependency upgrades, I'm fine with postponing this to a subsequent PR. The builds will be 50% longer, going from ~20 mins to ~30 mins. (And that 20 min number is already a 33% increase from ~15 mins).

src/test/java/io/confluent/idesidecar/restapi/integration/ConfluentPlatformIT.java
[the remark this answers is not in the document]
flippingbits: Sounds good! We want to make sure that (integration) tests always start with the same state so that they don't become flaky.

src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionState.java
```diff
@@ -183,4 +188,19 @@ public Optional<Credentials> getKafkaCredentials() {
   public Optional<Credentials> getSchemaRegistryCredentials() {
     return Optional.empty();
   }
+
+  /**
+   * Get the TLS configuration to use when connecting to the cluster (Kafka or Schema Registry). We
+   * assume that the TLS configuration is the same for both Kafka and Schema Registry. In the
```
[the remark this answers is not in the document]
rhauch: Actually, it's not clear whether or not SR supports mTLS for clients. That is an old issue that could have been not updated correctly. We need to verify with the Governance team, because the documentation does allude to it working.

[This section of the docs](https://docs.confluent.io/platform/current/security/authorization/rbac/configure-mtls-rbac.html#configure-sr) talks about how to enable mTLS in the Schema Registry config for CP. That's the server side config, and [cp-demo is definitely not configured to enable mTLS](https://github.com/confluentinc/cp-demo/blob/f0422a964fe8d111601df0d795fe774299da4b08/docker-compose.yml#L799-L804) since `ssl.client.authentication=NONE`, and it is slightly different than [this section that talks about configuring SR to use mTLS against Kafka](https://docs.confluent.io/platform/current/security/authentication/mutual-tls/overview.html#sr) (note the `kafkastore` prefix, which covers the configuration of Kafka clients that SR uses to talk with Kafka).

The [Schema Registry client configuration reference](https://docs.confluent.io/platform/current/schema-registry/sr-client-configs.html#ssl-keystore-location) shows this for `ssl.keystore.location` (emphasis mine):
> The location of the key store file. This is optional for client and can be used for *two-way authentication for client*.

The [Confluent Platform 7.8 announcement](https://www.confluent.io/blog/introducing-confluent-platform-7-8/#scale-platform-security-by-enabling-rbac-authorization-with-mtls-identity) says this (emphasis mine):
> With this enhancement, identities extracted from mTLS authentication can now be used to enforce RBAC authorization directly within the platform. This advancement allows customers to adopt RBAC without compromising their existing mTLS security framework, enabling both service-to-service authentication and granular access control. This is now **generally available across all Confluent Platform services** and includes supporting mTLS authentication to the Confluent Metadata Service.

[MDS clients](https://docs.confluent.io/platform/current/security/authorization/rbac/configure-mtls-rbac.html#configure-mds-client) can now also use mTLS.

The [release notes](https://docs.confluent.io/platform/current/release-notes/index.html#security) also says this under the "Security" section, but also this under the ["Schema Registry" section](https://docs.confluent.io/platform/current/release-notes/index.html#sr):
> RBAC can be enabled without requiring an identity provider by using mTLS identity for RBAC authorization for Schema Registry in Confluent Platform 7.8. For more information, see [Use mTLS Authentication with RBAC Authorization in Confluent Platform](https://docs.confluent.io/platform/current/security/authorization/rbac/mtls-rbac.html#mtls-rbac).

And finally, [this page](https://docs.confluent.io/platform/current/security/authorization/rbac/mtls-rbac.html#mtls-rbac) says:
> All the Confluent Platform REST interfaces accept TLS certificates for authentication from which the principal for RBAC authorization can be extracted from.

rhauch:
@rohitsanj wrote:
> Given the analysis above and the potential unknown impact beyond just a few dependency upgrades, I'm fine with postponing this to a subsequent PR. The builds will be 50% longer, going from ~20 mins to ~30 mins. (And that 20 min number is already a 33% increase from ~15 mins).

Thanks. We should continue to reduce the time required to run the CI builds, but now we just need it working and IMO we should separate out the improvements into subsequent PRs that can more easily be understood and rolled back if needed.

pom.xml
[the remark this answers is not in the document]
rhauch: Do you plan to make this change in this PR?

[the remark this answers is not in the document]
rohitsanj: I was planning on it, but I'll revert and make this change in a follow up.

src/main/java/io/confluent/idesidecar/restapi/clients/ClientConfigurator.java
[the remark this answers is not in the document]
rohitsanj: Removed

src/test/java/io/confluent/idesidecar/restapi/integration/ConfluentPlatformIT.java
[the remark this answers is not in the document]
rohitsanj: Changed it to shutdown by default.

rohitsanj:
@rhauch @flippingbits I've updated the PR description to reflect the current state of the changes. Please have another look! I apologize for the large file diff.

src/test/java/io/confluent/idesidecar/restapi/util/CPDemoTestEnvironment.java
rohitsanj: We can do this in a followup.

src/main/java/io/confluent/idesidecar/restapi/proxy/ConnectionProcessor.java
```diff
@@ -12,9 +12,9 @@
  * Processor that checks for the `x-connection-id` header and retrieves the connection state from
  * the connection state manager.
  *
- * @param <T> The proxy context type, which must extend {@link ProxyContext}
+ * @param <T> The proxy context type, which must extend {@link io.confluent.idesidecar.restapi.proxy.ClusterProxyContext}
  */
-public class ConnectionProcessor<T extends ProxyContext> extends
+public class ConnectionProcessor<T extends ClusterProxyContext> extends
```
rohitsanj, line 17: The generic type `T` needed to have context about the cluster type in order to set TLS configs appropriately.

This triggered a refactor to essentially move up all the fields from `ClusterProxyContext` into `ProxyContext`, and then rename `ProxyContext` to `ClusterProxyContext`.

src/main/java/io/confluent/idesidecar/restapi/credentials/TLSConfig.java
```diff
@@ -0,0 +1,375 @@
...
+        description = "Whether to verify the server certificate hostname."
+        + " Defaults to true if not set.",
+        defaultValue = DEFAULT_VERIFY_HOSTNAME_VALUE
+    )
+    @JsonProperty(value = "verify_hostname")
+    @Null
+    Boolean verifyHostname,
+
+    @Schema(
+        description = "Whether SSL is enabled. If not set, defaults to true.",
```
flippingbits, line 51: The second sentence is contradictory to requiring this field to be set.
I'd suggest to drop the constraint `required = true` and change `@NotNull` to `@Null`.

src/main/java/io/confluent/idesidecar/restapi/proxy/ConnectionProcessor.java
```diff
@@ -51,6 +51,25 @@ public Future<T> process(T context) {
...
+            });
+      case SCHEMA_REGISTRY ->
+        connectionState
+            .getSchemaRegistryTLSConfig()
+            .ifPresent(schemaRegistryOptions -> {
+              if (schemaRegistryOptions.truststore() != null) {
+                context.setTruststoreOptions(schemaRegistryOptions.truststore());
+              }
+            });
+    }
```
flippingbits, line 71: Do we need to provide the keystore in case of mTLS, too?

src/test/java/io/confluent/idesidecar/restapi/kafkarest/RecordsV3ErrorsSuite.java
```diff
@@ -274,6 +275,8 @@ default void shouldThrowNotImplementedForUnsupportedSchemaDetails(ProduceRequest
   }
 
   @Test
+  // TODO: Figure out why this test fails for cp-demo
+  @DisabledIfSystemProperty(named = "running-in-cp-test-environment", matches = "true")
```
flippingbits, line 279: Do we want to capture it in an issue?

src/test/java/io/confluent/idesidecar/restapi/kafkarest/RecordsV3WithoutSRSuite.java
```diff
@@ -1,6 +1,7 @@
 package io.confluent.idesidecar.restapi.kafkarest;
 
 import org.junit.jupiter.api.Test;
+import org.junitpioneer.jupiter.RetryingTest;
```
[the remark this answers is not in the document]
flippingbits: Should we remove it then?

src/test/java/io/confluent/idesidecar/restapi/resources/ConnectionsResourceTest.java
```diff
@@ -1280,7 +1276,99 @@ record TestInput(
...
+              "name": "Connection 1",
+              "type": "DIRECT",
+              "kafka_cluster": {
+                "bootstrap_servers": "localhost:9092"
+              },
+              "ssl": {
+                "truststore": {
+                  "path": "/path/to/truststore.jks"
+                }
+              }
```
flippingbits, line 1346: Should this be under `kafka_cluster`?

src/test/java/io/confluent/idesidecar/restapi/resources/ConnectionsResourceTest.java
```diff
@@ -1280,7 +1276,99 @@ record TestInput(
...
+              "ssl": {
+                "truststore": {
+                  "path": "/path/to/truststore.jks",
+                  "password": "truststore-password"
+                },
+                "keystore": {
+                  "path": "/path/to/keystore.jks",
+                  "password": "keystore-password",
+                  "key_password": "key-password"
+                }
```
flippingbits, line 1371: Same here.

src/test/java/io/confluent/idesidecar/restapi/resources/ConnectionsResourceTest.java
```diff
@@ -1349,6 +1437,73 @@ record TestInput(
...
+              }
+            }
+            """,
+            createError()
+                .withSource("kafka_cluster.ssl.truststore.path")
+                .withDetail("Kafka cluster truststore path is required and may not be blank")
+        ),
+        new TestInput(
+            "Direct spec is invalid with SSL having keystore only",
```
flippingbits, line 1461: Can't we rely on the built-in/system certs in the absence of the truststore?
AFAIK, that's how you'd [use mTLS with, for instance, Confluent Cloud](https://docs.confluent.io/cloud/current/security/authenticate/workload-identities/identity-providers/mtls/configure.html#configure-mtls).

flippingbits (commented):
Thanks for the updates, @rohitsanj! 🎉 I have a few questions and comments. Otherwise, your PR looks good to me. I feel you're very close to being able to merge it into `main`.

src/main/java/io/confluent/idesidecar/restapi/proxy/ConnectionProcessor.java
rohitsanj: I'm going to revert this change. Should substantially reduce the size of this PR's diff.

src/generated/resources/openapi.json
```diff
@@ -1166,6 +1163,42 @@
...
+        "required" : [ "path" ],
+        "type" : "object",
+        "properties" : {
+          "path" : {
+            "description" : "The path to the local key store file. Only specified if client needs to be authenticated by the server (mutual TLS).",
+            "maxLength" : 256,
+            "type" : "string"
+          },
+          "password" : {
+            "description" : "The password for the local key store file. If a password is not set, trust  store file configured will still be used, but integrity checking is disabled. A key store password is not supported for PEM format.",
```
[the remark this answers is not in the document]
rohitsanj: Yes, good catch, thanks.

src/generated/resources/openapi.json
```diff
@@ -1106,7 +1106,7 @@
       },
       "KafkaClusterConfig" : {
         "description" : "Kafka cluster configuration.",
-        "required" : [ "bootstrap_servers" ],
+        "required" : [ "bootstrap_servers", "ssl" ],
```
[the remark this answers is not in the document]
rohitsanj: Done, added `@Null` annotation as well as set `nullable = true` param in `@Schema` annotation.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -137,6 +139,34 @@ public Optional<Credentials> getSchemaRegistryCredentials() {
...
   }
 
+  @Override
+  public Optional<TLSConfig> getKafkaTLSConfig() {
+    if (spec.kafkaClusterConfig() != null) {
+      return Optional.of(
+          spec.kafkaClusterConfig().tlsConfig() != null
+          ? spec.kafkaClusterConfig().tlsConfig()
+          // Use the default TLS configuration if none is provided
+          : new TLSConfig()
```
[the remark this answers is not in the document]
rohitsanj: Definitely more readable and nicer overall, thanks!

src/test/java/io/confluent/idesidecar/restapi/kafkarest/RecordsV3ErrorsSuite.java
rohitsanj: Created https://github.com/confluentinc/ide-sidecar/issues/236

src/test/java/io/confluent/idesidecar/restapi/resources/ConnectionsResourceTest.java
rohitsanj: Yes, thanks for catching that. Seems we don't complain about unknown fields in the JSON payload.

src/test/java/io/confluent/idesidecar/restapi/resources/ConnectionsResourceTest.java
rohitsanj: Sounds good -- I'd imagine the built-in/system certs will be picked up automatically by the Java runtime without us having to configure anything.

flippingbits: We should definitely test if that works with the native executable.

rohitsanj: Sounds good -- I've already created https://github.com/confluentinc/ide-sidecar/issues/235 for testing mTLS against Confluent Cloud, testing would be taken care of as part of that.

rohitsanj: For now, I've changed the constraints to allow passing keystore options without truststore.

src/main/java/io/confluent/idesidecar/restapi/proxy/ConnectionProcessor.java
rohitsanj: Yes, but only for Schema Registry REST. We will think more about mTLS for Confluent Cloud as part of a follow up - #235

rohitsanj:
@flippingbits Thanks so much for the review! I've made some updates based on our offline conversation.

rohitsanj:
Seems there's quite a few branches in the ConnectionSpec validation logic that have not been covered with tests. Given how large this PR has gotten, we can add tests in a follow up PR -- https://github.com/confluentinc/ide-sidecar/issues/241

rohitsanj:
The two bugs reported by Sonarqube were false positives and I've resolved them with comments.

src/generated/resources/openapi.json
```diff
@@ -1166,6 +1164,42 @@
...
+        "required" : [ "path" ],
+        "type" : "object",
+        "properties" : {
+          "path" : {
+            "description" : "The path to the local key store file. Only specified if client needs to be authenticated by the server (mutual TLS).",
+            "maxLength" : 256,
+            "type" : "string"
+          },
+          "password" : {
+            "description" : "The password for the local key store file. If a password is not set, key  store file configured will still be used, but integrity checking is disabled. A key store password is not supported for PEM format.",
```
flippingbits, line 1177:
```suggestion
            "description" : "The password for the local key store file. If a password is not set, key store file configured will still be used, but integrity checking is disabled. A key store password is not supported for PEM format.",
```

src/generated/resources/openapi.json
```diff
@@ -1598,19 +1747,19 @@
...
-            "type" : "string"
-          },
           "data" : {
             "type" : "object",
             "nullable" : true
           },
+          "status" : {
+            "enum" : [ "UP", "DOWN" ],
+            "type" : "string"
+          },
```
flippingbits, line 1757: Do you mind removing this unrelated change from the PR?

flippingbits (commented):
Thanks for the update, @rohitsanj! This looks great. I have two more very minor comments. Otherwise, your PR looks good to me.

src/generated/resources/openapi.json
rohitsanj: Ah, good eye! I've fixed this.

src/generated/resources/openapi.json
rohitsanj: Removed, thanks!

flippingbits (approved):
Thanks for the great work, @rohitsanj! LGTM, pending the results of the CI pipeline.

rohitsanj:
For those following: @rhauch is fine with us merging this.

Issue #182, opened 2024-11-22
Set user-agent when interacting with CCloud


jlrobins:
Would be good for sidecar to interact with CCloud using a definite APIF team-approved User-Agent value that they will know how to aggregate.

rhauch:
The Confluent for VS Code extension should include the following User-Agent header value in CCloud requests:
```
Confluent-for-VSCode/v0.1.2 (https://confluent.io; support@confluent.io) sidecar/v0.10.20 (linux/amd64)
```

Changes to CCloud to recognize this agent are under way and should be in production in a few weeks, though those changes are not required and should not dictate the timing of this work.

Cerchie:
addressed via #202

PR #202, opened 2024-12-03
Cerchie/user agents

Cerchie:
Adds user agent in response to #182 

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -193,6 +249,7 @@ Stream<DynamicTest> testCombinations() {
               SidecarInfo sidecar = createSidecarInfo(input);
 
               // Verify the output matches
+              assert(input.userAgent).contains("support@confluent.io) sidecar/");
```
rohitsanj, line 252: Hmm, it seems we're not calling the newly added `userAgent()` method here and only asserting against the static value as provided by the test inputs. In order to properly test the method, I think this should instead be:

```suggestion
              assertEquals(input.userAgent, sidecar.userAgent())
```

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -28,7 +28,8 @@ record TestInputs(
       String osVersion,
       OperatingSystemType os,
       SidecarInfo.VsCode vscode,
-      String vscodeVersionPrefix
+      String vscodeVersionPrefix,
+      String userAgent
   ) {
```
rohitsanj, line 33: I believe we'll also need to specify the OS architecture as part of the test inputs, and pass it in the properties as `OS_ARCH_KEY` (like how we do for `OS_NAME_KEY` and `OS_VERSION_KEY`). This way, we override the OS arch of the machine the test itself is running on.

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -65,23 +68,32 @@ Stream<DynamicTest> testCombinations() {
...
             "Linux OS info with VS Code and version prefix",
             "Linux",
             "22.0413",
             OperatingSystemType.Linux,
             new SidecarInfo.VsCode("20.1.2", "1.2.3"),
-            "v"
+            "v",
+            "Confluent-for-VSCode/v%s (https://confluent.io; support@confluent.io) sidecar/v%s (Linux/x86_64)".formatted(
+                vscodeVersion,
+                SidecarInfo.VERSION)
```
rohitsanj, line 85: I think it would be better to statically specify what the output should be here rather than programmatically (or dynamically) specifying it. The reason is that we might end up exposing implementation details in our tests, and we ideally want to decouple the two. 

While writing this out, I realized that we might need to draw an exception for the tests-decoupled-from-implementation principle for baking in the `SidecarInfo.VERSION` into the expected user agent string. That said, it would be nice to be able to pass this into the `SidecarInfo` class rather than _always_ infer it from the `quarkus.application.version`. WDYT? cc @rhauch 

```suggestion
            "Confluent-for-VSCode/v20.1.2 (https://confluent.io; support@confluent.io) sidecar/v%s (Linux/x86_64)".formatted(SidecarInfo.VERSION)
```

src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java
```diff
@@ -1,14 +1,98 @@
...
+  @InjectSpy
+  SidecarInfo sidecarInfo;
+
+  @BeforeEach
+  public void setup() {
+    wireMockServer = new WireMockServer(options().dynamicPort());
+    wireMockServer.start();
+    WireMock.configureFor("localhost", wireMockServer.port());
+  }
+  @AfterEach
```
rohitsanj, line 41: nit: methods should be separated by one newline

```suggestion
  }

  @AfterEach
```

src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java
```diff
@@ -1,14 +1,98 @@
...
+  void resetWireMock() {
+    wireMockServer.stop();
+    wireMock.removeMappings();
+  }
+
+  @Test
+  void getDefaultWebClientOptionsIncludesUserAgent() {
+    var userAgent = webClientFactory.getDefaultWebClientOptions().getUserAgent();
+    Assertions.assertTrue(userAgent.contains("support@confluent.io) sidecar/"));
```
rohitsanj, line 50: Quarkus provides us a nice way of injecting specific application configuration overrides for test classes with the use of the [`@QuarkusTestProfile`](https://quarkus.io/guides/getting-started-testing#writing-a-profile) annotation. (Take a look at the [`NoAccessFilterProfile`](https://github.com/confluentinc/ide-sidecar/blob/main/src/test/java/io/confluent/idesidecar/restapi/testutil/NoAccessFilterProfile.java) and its usages for example). 

For this test, I believe we could hardcode some realistic values for the configuration that `SidecarInfo` expects (`vscode.version`, `os.version` and so on) by defining a new test profile class ([see StackOverflow](https://stackoverflow.com/a/69267852)). With that, we should be able to assert the expected value of _full_ user-agent string.

src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java
```diff
@@ -1,14 +1,98 @@
...
+                        .withBody("{\"message\":\"success\"}")
+                )
+    );
+
+    // Create a WebClient instance with the correct port
+    WebClient webClient = WebClient.create(
+        Vertx.vertx(), new WebClientOptions()
+            .setDefaultHost("localhost")
+            .setUserAgent("Confluent-for-VSCode/v%s (https://confluent.io; support@confluent.io) sidecar/v%s (%s/%s)"
+                .formatted(sidecarInfo.vsCode(),sidecarInfo.version(),sidecarInfo.osType(), sidecarInfo.osArch()))
```
rohitsanj, line 72: Like I mentioned in my other comment, we would ideally like to decouple the tests from the implementation being tested. Hence, we'd want to set the `WebClientOptions` as `webClientFactory.getDefaultWebClientOptions()` rather than constructing it here ourselves.

rohitsanj (changes requested):
This is a great start, @Cerchie! Thank you for adding the tests.

I've made some suggestions to tweak the implementation and the associated tests.

src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java
```diff
@@ -1,14 +1,98 @@
...
+            .setDefaultPort(wireMockServer.port()));
```
rhauch, line 73:
```suggestion
    WebClient webClient = WebClient.create(
        Vertx.vertx(), 
        new WebClientOptions()
            .setDefaultHost("localhost")
            .setUserAgent(
                "Confluent-for-VSCode/v%s (https://confluent.io; support@confluent.io) sidecar/v%s (%s/%s)".formatted(
                    sidecarInfo.vsCode()
                    sidecarInfo.version()
                    sidecarInfo.osType(),
                    sidecarInfo.osArch()
                )
            )
            .setDefaultPort(wireMockServer.port())
    );
```
And in fact you should be able to just do this:
```
WebClient webClient = webClientFactory.getWebClient();
webClient
    .getAbs(wireMockServer.baseUrl() + "/some-endpoint")
    .send()
 ...

**Review by rhauch — COMMENTED** [review]
I reviewed this with @Cerchie over zoom to help show how I review PRs, and because we'd discussed the general approach. I have a few comments, but GH issues prevented me from saving other changes.

**Comment by rhauch** [comment]
I should add that overall this is a great start, and for the most part the production code is just about right. Great job so far, @Cerchie!

File: src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java (hunk 7 above)
Comment (rohitsanj) [reply src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java:73]: `sidecarInfo.vsCode()` cannot be passed as-is since it returns an `Optional` instance. See my [comment](https://github.com/confluentinc/ide-sidecar/pull/202/files#r1868290337) above. That said, using `webClientFactory.getWebClient()` along with [my other recommendation](https://github.com/confluentinc/ide-sidecar/pull/202/files#r1868334337) to setup the environment should do the trick rather than constructing the user agent here.

File: src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java (hunk 5 above)
Comment (rhauch) [reply src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java:50]: @rohitsanj , IIUC test profiles work well for configurations that are already specified at build or runtime, like `vscode.version`, but it's not possible to specify values for system properties like `os.name`, etc. If my understanding is correct, then the test would still have to mock the `SidecarInfo` methods like `osType()` and `osArch()` anyway, so it'd be maybe simpler to also mock `vsCode()` rather than define a whole new test profile or even use an existing test profile and rely upon a combination of mock and test-profile configured values. Thoughts?

File: src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java (hunk 6 above)
Comment (rhauch) [reply src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java:72]: I suggested to use the factory, and mock some of the `SidecarInfo` methods used in `getUserAgent()` so that the test verifies that the factory creates a web client that includes the expected user agent. That, combined with the new `SidecarInfoTest` test methods that check that the `getUserAgent()` string is built correctly for different combinations of inputs, should probably be sufficient.

File: src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java (hunk 5 above)
Comment (rhauch) [reply src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java:50]: Oh, that's really good to know. Thanks for trying it out!

File: src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java [hunk 8]
```diff
@@ -58,186 +65,163 @@ public static SidecarInfo createSidecarInfo(TestInputs input) {
...
-                vscodeVersion,
-                SidecarInfo.VERSION)
+            "Confluent-for-VSCode/v1.2.3 (https://confluent.io; support@confluent.io) sidecar/v5.40.3 (linux/aarch64)"
         ),
         // Mac OS
         new TestInputs(
             "Mac OS info with VS Code",
-            "Mac OS X",
-            "13.1",
+            "mac os x",
```
rhauch, line 104: Maybe we should use `OperatingSystemType` enumeration literals here, because the literals don't have spaces.

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -58,186 +65,163 @@ public static SidecarInfo createSidecarInfo(TestInputs input) {
...
+            "x86_64",
+            "20.1.2",
             OperatingSystemType.MacOS,
             new SidecarInfo.VsCode("20.1.2", "1.2.3"),
             "",
-            "Confluent-for-VSCode/v%s (https://confluent.io; support@confluent.io) sidecar/v%s (Mac OS X/x86_64)".formatted(
-                vscodeVersion,
-                SidecarInfo.VERSION)
+            "Confluent-for-VSCode/v1.2.3 (https://confluent.io; support@confluent.io) sidecar/v5.40.3 (macosx/x86_64)"
```
rhauch, line 110: This should be:
```suggestion
            "Confluent-for-VSCode/v1.2.3 (https://confluent.io; support@confluent.io) sidecar/v5.40.3 (mac/x86_64)"
```
if we really are using `OperatingSystemType` literals.

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -58,186 +65,163 @@ public static SidecarInfo createSidecarInfo(TestInputs input) {
...
             "v",
-            "Confluent-for-VSCode/v%s (https://confluent.io; support@confluent.io) sidecar/v%s (Solaris 4/x86_64)".formatted(
-                vscodeVersion,
-                SidecarInfo.VERSION)
+            "Confluent-for-VSCode/v1.2.3 (https://confluent.io; support@confluent.io) sidecar/v5.40.3 (solaris/4.1.X.Y)"
         ),
         new TestInputs(
             "Solaris info without VS Code",
-            "Solaris 4",
+            "solaris",
```
rhauch, line 218: We don't have an enum literal for `SOLARIS`, or more correctly `UNIX`, in `OperatingSystemType`. If we want to distinguish, we should add the enum literal, and update the test.

src/test/java/io/confluent/idesidecar/restapi/testutil/FakeSideCarProfile.java
```diff
@@ -0,0 +1,21 @@
...
+public class FakeSideCarProfile implements QuarkusTestProfile {
+
+  /**
+   * Disable the access token filter to protest unit tests.
+   */
+  @Override
+  public Map<String, String> getConfigOverrides() {
+    return Map.of(
+        "vscode.version", "0.4.0 override",
+        "os.name", "linux-override",
```
rhauch, line 15: Is this representative of real `os.name` values? IIRC, most are space-separated with capitalized words.

src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java
```diff
@@ -64,41 +64,29 @@ public void webClientOptionsShouldContainUserAgentOnCreation() {
...
     );
 
-    // Create a WebClient instance with the correct port
-    WebClient webClient = WebClient.create(
-        Vertx.vertx(), new WebClientOptions()
-            .setDefaultHost("localhost")
-            .setUserAgent("Confluent-for-VSCode/v%s (https://confluent.io; support@confluent.io) sidecar/v%s (%s/%s)"
-                .formatted(sidecarInfo.vsCode(),sidecarInfo.version(),sidecarInfo.osType(), sidecarInfo.osArch()))
-            .setDefaultPort(wireMockServer.port()));
+    String userAgent = webClientFactory.getDefaultWebClientOptions().getUserAgent();
```
rhauch, line 67: Maybe we could also test that this is equivalent to `SidecarInfo.getUserAgent()`??

src/main/java/io/confluent/idesidecar/restapi/application/SidecarInfo.java
```diff
@@ -67,6 +67,7 @@ static String semanticVersionWithin(Properties props, String key, String def) {
 
   static final String OS_ARCH_KEY = "os.arch";
   static final String OS_NAME_KEY = "os.name";
+  static final String OS_TYPE_KEY = "os.type";
```
rhauch, line 70: `os.type` is not a real System property. We coopted the name `osType()` to be similar to `osName()` except that it returns one of the `OperatingSystemType` literals. See line #99 in this class.
```suggestion
```

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -34,8 +40,9 @@ record TestInputs(
     Map<String, String> asProperties() {
       // Construct the "system" properties from the input
       Map<String, String> props = new HashMap<>();
-      props.put(OS_NAME_KEY, osName);
-      props.put(OS_VERSION_KEY, osVersion);
+      props.put(OS_TYPE_KEY, osType);
```
rhauch, line 43: See my comment in `SidecarInfo`, re the `OS_TYPE_KEY` should not exist.
```suggestion
```

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -1,7 +1,10 @@
 package io.confluent.idesidecar.restapi.application;
 
+import static io.confluent.idesidecar.restapi.application.SidecarInfo.OS_ARCH_KEY;
 import static io.confluent.idesidecar.restapi.application.SidecarInfo.OS_NAME_KEY;
+import static io.confluent.idesidecar.restapi.application.SidecarInfo.OS_TYPE_KEY;
```
rhauch, line 5: See my comment in `SidecarInfo`, re the `OS_TYPE_KEY` should not exist.
```suggestion
```

rhauch (changes requested):
Thanks, @Cerchie. This is really close, but there still are a few minor details to address yet.

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
Cerchie: And this is because of 
```
    if (lowerName.contains("mac") || lowerName.contains("darwin")) {
      return OperatingSystemType.MacOS;
    }
```
in `OperatingSystemType` I imagine. The problem is sidecarinfo seems to return the uppercase type:

```
Expected :Confluent-for-VSCode/v1.2.3 (https://confluent.io; support@confluent.io) sidecar/v20.1.2 (linux/aarch64)
Actual   :Confluent-for-VSCode/v1.2.3 (https://confluent.io; support@confluent.io) sidecar/vunset (Linux/aarch64)
```

working on it.

Cerchie:
the tests in the individual files are passing now but I would say dubiously. Going to explain why in the comments

src/main/java/io/confluent/idesidecar/restapi/util/OperatingSystemType.java
```diff
@@ -6,6 +6,7 @@ public enum OperatingSystemType {
   Windows,
   MacOS,
   Linux,
+  Solaris,
```
Cerchie, line 9: added Solaris for more test granularity.

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -57,131 +64,162 @@ public static SidecarInfo createSidecarInfo(TestInputs input) {
...
-            "Linux",
-            "22.0413",
+            "linux",
+            "aarch64",
+            "2.1.2",
             OperatingSystemType.Linux,
             new SidecarInfo.VsCode("20.1.2", "1.2.3"),
-            null
+            null,
+            "Confluent-for-VSCode/v1.2.3 (https://confluent.io; support@confluent.io) sidecar/vunset (Linux/aarch64)"
```
Cerchie, line 78: These are coming in from the sidecar capitalized, although we're setting os type in there.

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -25,16 +29,19 @@ public class SidecarInfoTest {
...
       OperatingSystemType os,
       SidecarInfo.VsCode vscode,
-      String vscodeVersionPrefix
+      String vscodeVersionPrefix,
+      String userAgent
   ) {
     Map<String, String> asProperties() {
       // Construct the "system" properties from the input
       Map<String, String> props = new HashMap<>();
+      props.put(OS_ARCH_KEY, osArch);
```
Cerchie, line 42: we set the osArch here, but you'll notice it always comes in as the same value in the tests. ("aarch64")

src/main/java/io/confluent/idesidecar/restapi/application/SidecarInfo.java
```diff
@@ -148,4 +157,13 @@ static String getSystemOrEnvProperty(String name) {
     var result = System.getProperty(name);
     return result != null ? result : System.getenv(name);
   }
+
+  public String getUserAgent() {
+   return "Confluent-for-VSCode/v%s (https://confluent.io; support@confluent.io) sidecar/v%s (%s/%s)".formatted(
+        vsCode().map(VsCode::extensionVersion).orElse("unknown"),
+        version(),
```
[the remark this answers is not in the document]
rohitsanj: The [`quarkus.application.version`](https://quarkus.io/guides/all-config#quarkus-core_quarkus-application-version) is automatically set by Quarkus based on the version specified in the `pom.xml` file.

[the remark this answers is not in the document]
Cerchie: ok, so this should work regardless?

src/main/java/io/confluent/idesidecar/restapi/util/OperatingSystemType.java
rhauch: Do we want `Solaris`, or `Unix` since Solaris is one of many Unixes?

src/main/java/io/confluent/idesidecar/restapi/application/SidecarInfo.java
```diff
@@ -148,4 +157,13 @@ static String getSystemOrEnvProperty(String name) {
...
+        osType(),
```
rhauch, line 165: I think we need to lowercase the `OperatingSystemType` here:
```suggestion
        osType().name().toLowerCase(),
```

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
rhauch: Right, but we should expect the OS type to be lowercased. See my comment in this review that suggests changes in `SidecarInfo.getUserAgent()`.

src/main/java/io/confluent/idesidecar/restapi/application/SidecarInfo.java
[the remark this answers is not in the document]
rohitsanj: Yes, it should just work :+1:

Cerchie:
Summary of changes made in response to reviews: 
- addition of `FakeSideCarProfile` test class for use in testing the web client factory 
- dynamic inputs for sidecar version and operating system architecture so tests can respond to changes properly; even with the `TestInput` class the sidecar instance picks these up dynamically 
- using a `contains` for verifying the presence of the user agent in the default webclientoptions 
- general cleanup & details pointed out in reviews

Note: I kept in the 'override' bits of string in the `FakeSideCar` in order to signal where these values were coming from. Happy to make changes to make these more realistic if needed, but I thought it'd be helpful since I myself had trouble keeping track of this while writing the tests, and it might help a reader.

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -54,134 +59,212 @@ public static SidecarInfo createSidecarInfo(TestInputs input) {
...
-            OperatingSystemType.Other,
+            "20.1.2",
+            OperatingSystemType.Unix,
             new SidecarInfo.VsCode("20.1.2", "1.2.3"),
-            ""
+            "",
+            "Confluent-for-VSCode/v1.2.3 (https://confluent.io; support@confluent.io) sidecar/v%s (unix/%s)"
+                .formatted(
+                    sidecarInfo.version(),
+                    sidecarInfo.osArch())
```
rohitsanj, line 241: Hmm, I think we shouldn't use the `sidecarInfo` instance to construct the expected output string (I alluded to this in my [earlier comment](https://github.com/confluentinc/ide-sidecar/pull/202#discussion_r1868306829)). By doing so, we'd be "baking in" the implementation into the tests. 

If we were to modify the implementation to return something other than a `String` from the `version()` or `osArch()` methods, we should expect these tests to break. However, since we're using the source implementation itself to construct the expected value, these tests would continue to pass. 

```suggestion
            "Confluent-for-VSCode/v1.2.3 (https://confluent.io; support@confluent.io) sidecar/v20.1.2 (unix/aarch64)"
```

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -193,9 +276,8 @@ Stream<DynamicTest> testCombinations() {
               SidecarInfo sidecar = createSidecarInfo(input);
 
               // Verify the output matches
+              assertEquals(input.userAgent, sidecar.getUserAgent());
               assertEquals(input.os, sidecar.osType());
-              assertEquals(input.osName, sidecar.osName());
```
rohitsanj, line 197: Why did we remove this assertion?

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -54,134 +59,212 @@ public static SidecarInfo createSidecarInfo(TestInputs input) {
...
+                    sidecarInfo.version(),
+                    sidecarInfo.osArch())
         ),
         // Other
         new TestInputs(
-            "Solaris info with VS Code",
-            "Solaris 4",
+            "Unix info with VS Code",
+            "unix",
             "4.1.X.Y",
```
rohitsanj, line 233: Based on the `TestInput` record members, the third argument corresponds to the `osArch` and not `osVersion`. This needs to be updated to reflect a valid osArch.

src/test/java/io/confluent/idesidecar/restapi/testutil/FakeSideCarProfile.java
```diff
@@ -0,0 +1,22 @@
...
+import io.quarkus.test.junit.QuarkusTestProfile;
+import java.util.Map;
+
+public class FakeSideCarProfile implements QuarkusTestProfile {
+
+  /**
+   * Disable the access token filter to protest unit tests.
+   */
+  @Override
+  public Map<String, String> getConfigOverrides() {
```
rohitsanj, line 12: These values need to be updated to reflect real values, the "override" values I showcased in my screenshot was just for demonstration purposes.

src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java
rohitsanj: We should assert this as Randall mentions. 

We could possibly pass in a different user agent to the WebClient and we'd want this test to fail if that happens so we can investigate to see if that's the desired behavior.

rohitsanj (changes requested):
@Cerchie, thank you for the updates! I think the PR is very close, I do have a few suggestions on improving the tests.

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -63,125 +70,163 @@ Stream<DynamicTest> testCombinations() {
...
             "Mac OS info without VS Code",
             "Mac OS X",
             "13.1",
+            "amd64",
+            "1.150.2",
             OperatingSystemType.MacOS,
             null,
-            null
-        ),
-        // Windows 10
```
rohitsanj, line 111: Let's revert this whitespace diff

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -63,125 +70,163 @@ Stream<DynamicTest> testCombinations() {
...
+            "x86_64",
+            "1.150.2",
             OperatingSystemType.Windows,
             null,
-            null
+            null,
+            "Confluent-for-VSCode/vunknown (https://confluent.io; support@confluent.io) sidecar/v1.150.2 (windows/x86_64)"
         ),
-        // Windows 11
+ // Windows 11
```
rohitsanj, line 164: Let's revert this whitespace diff

src/test/java/io/confluent/idesidecar/restapi/testutil/FakeSideCarProfile.java
```diff
@@ -0,0 +1,22 @@
...
+    return Map.of(
+        "vscode.version", "0.4.0 override",
+        "os.name", "linux-override",
+        "os.arch", "amd64-override",
+        "os.version", "5.11.0-27-azure-override",
+        "version", "0.105.0 override",
+        "vscode.extension.version", "0.21.3 override"
+    );
```
rohitsanj, line 20: Let's use some real values here instead as follows:

```suggestion
  public Map<String, String> getConfigOverrides() {
    return Map.of(
        "vscode.version", "20.1.2",
        "os.name", "GNU/Linux",
        "os.arch", "amd64",
        "os.version", "5.15.0-60-generic",
        "quarkus.application.version", "1.2.3",
        "vscode.extension.version", "0.21.0"
    );
```

Notice that I've renamed `version` -> `quarkus.application.version` since that's the property we lookup in the `SidecarInfo` class.

src/test/java/io/confluent/idesidecar/restapi/util/WebClientFactoryTest.java
```diff
@@ -1,20 +1,89 @@
...
+    wireMockServer.stubFor(
+        WireMock.get(urlPattern)
+                .willReturn(
+                    aResponse()
+                        .withStatus(200)
+                        .withBody("{\"message\":\"success\"}")
+                )
+    );
+
+    String userAgent = webClientFactory.getDefaultWebClientOptions().getUserAgent();
```
rohitsanj, line 59: Let's add this assertion here as follows (as mentioned by Randall [here](https://github.com/confluentinc/ide-sidecar/pull/202/files#r1870476516))

```suggestion
    String userAgent = webClientFactory.getDefaultWebClientOptions().getUserAgent();

    // Assert that this matches what SidecarInfo tells us the user agent is
    assertEquals(userAgent, sidecarInfo.userAgent());
```

rohitsanj (changes requested):
Thanks again, @Cerchie. I've recommended a few more changes.

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -63,125 +70,163 @@ Stream<DynamicTest> testCombinations() {
...
+            "amd64",
+            "1.150.2",
             OperatingSystemType.MacOS,
             null,
-            null
-        ),
-        // Windows 10
+            null,
+            "Confluent-for-VSCode/vunknown (https://confluent.io; support@confluent.io) sidecar/v1.150.2 (macos/amd64)"),
+   // Windows 10
```
rohitsanj, line 131: Let's align this comment against the `new` keyword in the following line.

src/main/java/io/confluent/idesidecar/restapi/application/SidecarInfo.java
```diff
@@ -133,6 +134,8 @@ public Optional<VsCode> vsCode() {
     return vscode;
   }
 
+  public String osArch() { return osArch; }
```
rohitsanj, line 137: Ah, sorry, I just noticed that this is a one-liner. Let's please split it up like so:
```suggestion
  public String osArch() {
    return osArch;
  }
```

src/main/java/io/confluent/idesidecar/restapi/util/OperatingSystemType.java
```diff
@@ -24,6 +25,10 @@ public static OperatingSystemType from(Properties system) {
     if (lowerName.contains("mac") || lowerName.contains("darwin")) {
       return OperatingSystemType.MacOS;
     }
+    if (lowerName.contains("unix" ) || lowerName.contains("sunos") || lowerName.contains("solaris")
```
rohitsanj, line 28: nit: Let's remove this trailing whitespace.
```suggestion
    if (lowerName.contains("unix") || lowerName.contains("sunos") || lowerName.contains("solaris")
```

rohitsanj (changes requested):
Just a couple more that I missed initially, apologies.

rohitsanj (approved):
LGTM, thanks! Let's wait for @rhauch to approve as well.

rhauch (approved):
LGTM. Nice work, @Cerchie!

Issue #128, opened 2024-10-26
Add support for connecting to and using Confluent Platform MDS


rhauch:
Implement the existing `PLATFORM` connection type (or rename to `MDS`) that connects to an MDS URL using basic credentials (see #124) and uses the metadata API to dynamically discover the available Kafka and SR clusters that were registered with MDS. Once connected, the GraphQL response will include these clusters. Like with direct connections, these connections should provide configuration property overrides for the the Kafka and SR clients.

PR #205, opened 2024-12-05
Add initial partial support for connecting to Confluent Platform MDS

rhauch:
Part one of #128 

## Summary of Changes

This is the first part of the implementation to allow connections to Confluent Platform MDS, which contains a registry of one or more Kafka clusters and one or more Schema Registry clusters. (There may be other types of clusters in the registry, too, but the VS Code extension only recognizes and uses Kafka and SR clusters.)

Once #128 is complete, these clusters will visible as resources in GraphQL query responses, and will allow the VS Code extension to navigate topics and schemas, consume and produce messages, and otherwise use the Kafka REST API and SR APIs on these Kafka and SR clusters (respectively).

This PR adds some of this functionality. It is possible to create/update a connection to an MDS, and the status of the connection will correctly reflect the ability to authenticate using the supplied credentials to MDS. This PR builds upon the nascent `type: PLATFORM` connection type that before now did basically nothing. It adds an `spec.mds` object to the Connection resource to specify the MDS URL and credentials. On `POST` and `PUT` calls to the Connection API, validation is performed on all the individual fields; invalid fields will result in an HTTP error. If successful, the connection's status will be asynchronously updated, just like other connections, to reflect whether the URL and credentials were sufficient to allow successfully authenticating to MDS.

As of this PR, the functionality does not yet:
* fetch cluster resources (GraphQL)
* add many tests; ITs for MDS connections will need #201

A few dynamic tests were removed, since they relied upon platform connections doing nothing.

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [x] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


rohitsanj:
Closing this PR given the following decision (facilitated by @flippingbits):

> We’ll support Confluent Platform resources via Direct connections in the near term and consider offering discovery capabilities on top of MDS at a later point.

PR #209, opened 2024-12-06
Changes the DirectConnectionState to use the default configs for AdminClients

### LINKED ISSUE #207: Use default admin client configs in `DirectConnectionState`
The `DirectConnectionState` [creates its own AdminClient](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java#L288). We should make sure that it picks up the [default admin client config](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/resources/application.yml#L151) from the `application.yml`.

rhauch:
Fixes #207

## Summary of Changes

Changes the `ClientConfiguration` so that the static methods do use the default configurations for admin clients, consumers, and producers.

Note that we need the static methods in `ClientConfiguration` because `DirectConnectionState` needs them (and `PlatformConnectionState` will, too), and `ConnectionState` objects are not beans (and therefore can’t inject the `ClientConfigurator` bean) and are created as normal objects in many of the tests. But we still want the `ClientConfigurator` bean methods to find the connections from the `ConnectionManager` bean and the clusters in the `ClusterCache` bean, while the static methods (that do the real work) require the `ConnectionState` object to be passed in as parameters.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


flippingbits (approved):
LGTM! Thanks, @rhauch.

PR #214, opened 2024-12-10
revert change to version var

Cerchie:
## Summary of Changes

vscode release is [blocked ](https://github.com/confluentinc/vscode/pull/730)by latest [change](https://github.com/confluentinc/ide-sidecar/pull/202/files#diff-8715e6c09c7231caa991d914500e23067980a88baf188deb940dd32fbb45fb60R65) to SideCarInfo for user agents. This reverts that change. 

## Pull request checklist

- Tests:
    - [ ] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/application/SidecarInfo.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/application/SidecarInfo.java
@@ -63,10 +63,10 @@ public class SidecarInfo {
   static final String OS_ARCH_KEY = "os.arch";
   static final String OS_NAME_KEY = "os.name";
   static final String OS_VERSION_KEY = "os.version";
-  static final String VERSION = ConfigProvider
+  static final String SIDECAR_VERSION_KEY = ConfigProvider
       .getConfig()
       .getOptionalValue("quarkus.application.version", String.class)
-      .orElse(UNSET_VERSION);;
+      .orElse(UNSET_VERSION);
   static final String VSCODE_VERSION_ENV = "VSCODE_VERSION";
   static final String VSCODE_VERSION_KEY = "vscode.version";
   static final String VSCODE_EXTENSION_VERSION_ENV = "VSCODE_EXTENSION_VERSION";
@@ -94,7 +94,7 @@ public class SidecarInfo {
     // Get the OS information
     osName = system.getProperty(OS_NAME_KEY, "unknown");
     osVersion = system.getProperty(OS_VERSION_KEY, "unknown");
-    sidecarVersion =  system.getProperty(VERSION, UNSET_VERSION);
+    sidecarVersion =  system.getProperty(SIDECAR_VERSION_KEY, UNSET_VERSION);
     osArch =  system.getProperty(OS_ARCH_KEY, "unknown");
 
     // Determine the best-matching OS type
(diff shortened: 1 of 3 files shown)


src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
```diff
@@ -41,7 +41,7 @@ Map<String, String> asProperties() {
       props.put(OS_ARCH_KEY, osArch);
       props.put(OS_NAME_KEY, osName);
       props.put(OS_VERSION_KEY, osVersion);
-      props.put(SIDECAR_VERSION_KEY, sidecarVersion);
+      props.put(VERSION, sidecarVersion);
```
flippingbits, line 44: I think we should keep `SIDECAR_VERSION_KEY` here.

src/main/java/io/confluent/idesidecar/restapi/application/SidecarInfo.java
```diff
@@ -90,7 +94,7 @@ public SidecarInfo() {
     // Get the OS information
     osName = system.getProperty(OS_NAME_KEY, "unknown");
     osVersion = system.getProperty(OS_VERSION_KEY, "unknown");
-    sidecarVersion =  system.getProperty(SIDECAR_VERSION_KEY, UNSET_VERSION);
+    sidecarVersion =  system.getProperty(VERSION, UNSET_VERSION);
```
flippingbits, line 97:
```suggestion
    sidecarVersion =  VERSION;
```

flippingbits (commented):
Thanks, @Cerchie! I have three minor questions/suggestions.

src/test/java/io/confluent/idesidecar/restapi/application/SidecarInfoTest.java
Cerchie: yeah makes sense it's helpful to separate from all the other versions in the file

src/main/java/io/confluent/idesidecar/restapi/application/SidecarInfo.java
```diff
@@ -62,7 +63,10 @@ static String semanticVersionWithin(Properties props, String key, String def) {
   static final String OS_ARCH_KEY = "os.arch";
   static final String OS_NAME_KEY = "os.name";
   static final String OS_VERSION_KEY = "os.version";
-  static final String SIDECAR_VERSION_KEY = "quarkus.application.version";
+  static final String SIDECAR_VERSION_KEY = ConfigProvider
```
flippingbits, line 66:
```suggestion
  static final String SIDECAR_VERSION = ConfigProvider
```

Also, can you remove the attribute `sidecarVersion` and let the method `version()` return the constant variable `SIDECAR_VARIABLE`?

flippingbits (commented):
Thanks for the update, @Cerchie! I have one more question.

src/main/java/io/confluent/idesidecar/restapi/application/SidecarInfo.java
Cerchie: This breaks the tests:

```
      assertEquals(input.osVersion, sidecar.osVersion());
```

I'm torn between removing this from the tests and leaving it in... in case another part of the codebase will eventually rely on this attribute

flippingbits: Hm, I don't think it will break this assertion but you need to update some other assertions because we can no longer pass a static sidecar version in the tests. Let's discuss it offline.

flippingbits (approved):
Thanks for the updates. LGTM, @Cerchie!

PR #221, opened 2024-12-16
Suggested changes to #215

rhauch:
Some suggestions for #215, including:
- Use formatted logging in remaining places.
- Use record types as much as possible for WS messages, and adopt getter functions in record types without `get`
- Use Jackson’s Deduction feature to figure out the correct implementation of the subclass during deserialization, defaulting to DynamicMessageBody
- Changed message type to enum rather than string, for stronger typing, with custom deserialization logic to support UNKNOWN (and case-insensitive while we’re at it)
- Fixed imports and package qualifications that weren’t needed


jlrobins (approved):
Thanks! More java-y!

PR #222, opened 2024-12-16
Broadcast connection-related events

rhauch:
## Summary of Changes

Add a notional WebSocket message type for signaling connection-related changes, such as when connections are created, updated, deleted, or when their status changes.

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/main/java/io/confluent/idesidecar/websocket/messages/ConnectionEventBody.java
```diff
@@ -0,0 +1,31 @@
+package io.confluent.idesidecar.websocket.messages;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.confluent.idesidecar.restapi.connections.ConnectionState;
+import io.confluent.idesidecar.restapi.models.ConnectionSpec.ConnectionType;
+import io.confluent.idesidecar.restapi.models.ConnectionStatus;
+
+/**
+ * Sent by the sidecar to the workspace when a connection is created, when existing connections
```
flippingbits, line 9:
```suggestion
 * Sent by the sidecar to the workspaces when a connection is created, when existing connections
```

src/main/java/io/confluent/idesidecar/websocket/messages/ConnectionEventBody.java
```diff
@@ -0,0 +1,31 @@
...
+ * are updated or their status changed, or when a connection is deleted.
+ */
+public record ConnectionEventBody(
+    @JsonProperty("connection_id") String connectionId,
+    @JsonProperty("connection_type") ConnectionType connectionType,
+    @JsonProperty("event_kind") EventKind eventKind,
+    @JsonProperty("connection_status")ConnectionStatus connectionStatus
```
flippingbits, line 16:
```suggestion
    @JsonProperty("connection_status") ConnectionStatus connectionStatus
```

flippingbits, line 16: What about adding the entire [Connection](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/models/Connection.java) model, including the spec, to the event instead of sending the `connectionId`, `connectionType`, and `connectionStatus` as separate fields?

Also, do we want to add a timestamp field that holds the time when the event has been created?

flippingbits (commented):
Great job, @rhauch! I wanted to discuss two questions with you. Otherwise, your PR looks good to me.

src/main/java/io/confluent/idesidecar/websocket/messages/ConnectionEventBody.java
rhauch: I didn't add a timestamp or sequence number. Let's add those if/when needed.

Cerchie (dismissed):
mostly reviewing to catch up on this change, but I can verify the test file is passing :) 

nit: this method was not introduced in the PR's changes but I noticed the use of `thread.sleep` in WebsocketEndpointTest. Could be an opportunity to swap out for Awaitility.

rhauch:
The recently-added WebSocket unit tests seem to be a bit flaky, with `WebsocketEndpointTest.testWebsocketLifecycle()` timing out in several different places. I've upped the timeout, just in case we're close to the limits.

flippingbits (approved):
Thanks for the update, @rhauch! I have one minor question. Otherwise, your PR looks good to me.

I'm already approving the PR in case you decide against updating the log statement.

rhauch:
@flippingbits, I fixed the log messages to include the exception, per your suggestion. Mind taking another look?

flippingbits (approved):
LGTM! Unfortunately, it looks like the flaky test has failed again.

rhauch:
Superseded by #243

rhauch:
Superseded by #243

PR #224, opened 2024-12-17
Fix coverage reporting and push to Sonarqube

rohitsanj:
### Coverage reporting related changes made to `pom.xml`

- Sets `quarkus.test.arg-line` to the JaCoCo Java agent in the maven-failsafe-plugin configuration. This lets JaCoCo track execution data from the integration tests which was previously not being picked up.
- Add `executions` in the JaCoCo plugin configuration for: `prepare-agent-integration` (this sets `jacoco.failsafe.args`), `report-it` (generates report for integration tests), `merge-results` (merges unit test and integration test execution data), `report-merged` (generates combined report for unit + int tests)
  - Also, renamed existing `generate-code-coverage-report` to `report-unit-tests`. Removed `report-aggregate` goal since it is only applicable for multi module projects.
- Added a new `prepare-build-integration-test` execution to the `quarkus-maven-plugin` configuration -- this sets the system property `quarkus.package.write-transformed-bytecode-to-build-output=true` which is [required](https://quarkus.io/guides/tests-with-coverage#coverage-for-integration-tests) for correctly reporting coverage data from integration tests.
- Removed the separate reporting configuration for `jacoco-maven-plugin` -- again, this is only useful for multi module projects.

### Sonarqube related changes

- Configured `sonarqube` in service.yml (also changed sonar.properties by hand, this file would be updated automatically to the same contents by ServiceBot once this PR is merged)
- Added the `emit-sonarqube-data` script invocation to semaphore.yml

### Misc changes

- Set `<id>integration-test</id>` in the default maven-failsafe-plugin configuration to prevent integration tests from running twice with the `native` profile.

#231 must be merged before this PR.


pom.xml
```diff
@@ -937,24 +1003,6 @@
           </reportSet>
         </reportSets>
       </plugin>
-      <plugin>
```
rohitsanj, line 940: We don't need to run the `report-aggregate` goal since ide-sidecar is a single module project.

.semaphore/semaphore.yml
```diff
@@ -34,23 +34,35 @@ global_job_config:
...
+  - name: "Build JARs and Unit Test"
     dependencies: []
     task:
       prologue:
         commands:
           - sem-version java 21
+      epilogue:
+        always:
+          commands:
+            - emit-sonarqube-data --run_only_sonar_scan
```
rohitsanj, line 46: We split the "Build and Test" block into two so that we can configure `epilogue.always` on just "Build JARs and Unit Test"

sonar-project.properties
```diff
@@ -1,8 +1,8 @@
 ### service-bot sonarqube plugin managed file
 sonar.coverage.exclusions=**/test/**/*,**/tests/**/*,**/mock/**/*,**/mocks/**/*,**/*mock*,**/*test*
-sonar.coverage.jacoco.xmlReportPaths=**/jacoco.xml
+sonar.coverage.jacoco.xmlReportPaths=**/jacoco-report/jacoco.xml,**/jacoco-report-it/jacoco.xml
 sonar.cpd.exclusions=**/test/**/*,**/tests/**/*,**/mock/**/*,**/mocks/**/*,**/*mock*,**/*test*
-sonar.exclusions=**/*.pb.*,**/mk-files/**/*,**/test/**/*
+sonar.exclusions=**/*.pb.*,**/mk-files/**/*,**/test/**/*,**/static/templates/**/*
 sonar.java.binaries=.
```
rohitsanj, line 6: These changes would be made by ServiceBot given the `service.yml` changes, but I wanted it to take effect in this PR.

flippingbits (approved):
Thanks, @rohitsanj. This is huge. LGTM!

rohitsanj:
Huge? Wait till you see #201. /s

PR #225, opened 2024-12-17
MINOR(OpenAPI): remove Credentials reference altogether

shouples:
## Summary of Changes

Context: We have a type issue in the VS Code extension client code generation where the `allOf` pointing to `Credentials` (of type `object`) is chosen instead of the `oneOf: BasicCredentials | ApiKeyAndSecret`.

This PR aims to tighten up the sections for `credentials` under the `KafkaClusterConfig` and `SchemaRegistryConfig` so we don't mix `allOf` (with a basic `object` type) and `oneOf`.

#### Current (`main`):
https://github.com/confluentinc/ide-sidecar/blob/9203449cce9f38a236f2c473d314b01ed6017802/src/generated/resources/openapi.yaml#L813-L822

#### New:
https://github.com/confluentinc/ide-sidecar/blob/b8c02495ef797fcfd22cb1de5fdf2cb5e5e3c90a/src/generated/resources/openapi.yaml#L811-L817

For future use: setting `implementation = Object.class` (for a `Schema` annotation on a property) will prevent another property's interface from being included in that section of the OpenAPI spec. https://download.eclipse.org/microprofile/microprofile-2.0-javadocs-test/apidocs/org/eclipse/microprofile/openapi/annotations/media/Schema.html#implementation--

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/generated/resources/openapi.json
+++ b/src/generated/resources/openapi.json
@@ -965,9 +965,6 @@
           }
         }
       },
-      "Credentials" : {
-        "type" : "object"
-      },
       "Error" : {
         "description" : "Describes a particular error encountered while performing an operation.",
         "type" : "object",
@@ -1120,10 +1117,6 @@
           },
           "credentials" : {
             "description" : "The credentials for the Kafka cluster, or null if no authentication is required",
-            "type" : "object",
-            "allOf" : [ {
-              "$ref" : "#/components/schemas/Credentials"
-            } ],
             "oneOf" : [ {
               "$ref" : "#/components/schemas/BasicCredentials"
             }, {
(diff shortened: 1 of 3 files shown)


rhauch:
All other things being equal, I personally like this approach better than #226 because it much more easily allows us to control/limit which `Credentials` subtypes are allowed in each of the places it's used. Currently we _do_ happen to allow all `Credentials` subtypes to be used in both `SchemaRegistryConfig` but do want it in `KafkaClusterConfig`, but maybe that is not the case in the future, either because we add a new credential type that works in one but not the other, or we add a new system (e.g., Flink) that supports a different set of credentials.

rhauch (commented):
Let's just revert the changes to the health check and response types, and otherwise this is perfect. Thanks for figuring out how to get the OpenAPI spec generator to bend to your wishes!

rhauch (approved):
LGTM. Thanks for the fix!

PR #226, opened 2024-12-17
MINOR(OpenAPI): remove `oneOf` in favor of Credentials reference

shouples:
## Summary of Changes

Context: We have a type issue in the VS Code extension client code generation where the `allOf` pointing to `Credentials` (of type `object`) is chosen instead of the `oneOf: BasicCredentials | ApiKeyAndSecret`.

This PR aims to tighten up the sections for `credentials` under the `KafkaClusterConfig` and `SchemaRegistryConfig` so we don't mix `allOf` (with a basic `object` type) and `oneOf`. 

#### Current (`main`):
https://github.com/confluentinc/ide-sidecar/blob/9203449cce9f38a236f2c473d314b01ed6017802/src/generated/resources/openapi.yaml#L813-L822

#### New:
https://github.com/confluentinc/ide-sidecar/blob/4cf724ef14e457bb02c41f78ffa969bd576c0c62/src/generated/resources/openapi.yaml#L819-L825

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/generated/resources/openapi.yaml
+++ b/src/generated/resources/openapi.yaml
@@ -694,7 +694,13 @@ components:
           items:
             $ref: "#/components/schemas/Connection"
     Credentials:
+      description: "The credentials for a ConnectionSpec config, or null if no authentication\
+        \ is required"
       type: object
+      oneOf:
+      - $ref: "#/components/schemas/BasicCredentials"
+      - $ref: "#/components/schemas/ApiKeyAndSecret"
+      nullable: true
     Error:
       description: Describes a particular error encountered while performing an operation.
       type: object
@@ -811,15 +817,7 @@ components:
           minLength: 1
           type: string
         credentials:
-          description: "The credentials for the Kafka cluster, or null if no authentication\
-            \ is required"
-          type: object
-          allOf:
-          - $ref: "#/components/schemas/Credentials"
-          oneOf:
-          - $ref: "#/components/schemas/BasicCredentials"
-          - $ref: "#/components/schemas/ApiKeyAndSecret"
-          nullable: true
+          $ref: "#/components/schemas/Credentials"
         ssl:
           description: "Whether to communicate with the Kafka cluster over TLS/SSL.\
             \ Defaults to 'true', but set to 'false' when the Kafka cluster does not\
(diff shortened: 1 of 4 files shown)


shouples:
If this PR is approved, we should close https://github.com/confluentinc/ide-sidecar/pull/225

src/generated/resources/openapi.yaml
```diff
@@ -694,7 +694,13 @@ components:
           items:
             $ref: "#/components/schemas/Connection"
     Credentials:
+      description: "The credentials for a ConnectionSpec config, or null if no authentication\
```
shouples, line 697: Could stand to be reworded if not removed altogether?

src/generated/resources/openapi.yaml
```diff
@@ -1188,13 +1188,13 @@ components:
...
+        name:
           type: string
         data:
           type: object
           nullable: true
-        name:
+        status:
+          enum:
+          - UP
+          - DOWN
```
shouples, line 1199: Unrelated to this PR. No idea why this shuffles around

rhauch: The OpenAPI spec generator does not deterministically generate the health check or response types, and will alternate as other changes are made in the codebase.

I would just manually exclude those line changes from the PR. I often will add line changes manually for the `openapi.yml` and `openapi.json` files, excluding any health-related changes (unless that's intentional), rather than just adding all changes in those files.

rhauch:
All other things being equal, I personally like #225 approach better than this approach, because #225 much more easily allows us to specify in each place `Credentials` is used the _potentially different_ subset of `Credentials` subtypes. Currently we _do_ happen to allow all `Credentials` subtypes to be used in both `SchemaRegistryConfig` but do want it in `KafkaClusterConfig`, but maybe that is not the case in the future, either because we add a new credential type that works in one but not the other, or we add a new system (e.g., Flink) that supports a different set of credentials.

See https://github.com/confluentinc/ide-sidecar/pull/225#issuecomment-2549478361

shouples:
Closing in favor of #225

PR #227, opened 2024-12-17
Minor improvements to the WebsocketEndpoint

rhauch:
Suggested fixes for #215 

Mostly formatting and line lengths, but also refactored the `sendErrorAndCloseSession(…)` method to format and log the message, and better handle errors while ensuring the session is closed.


jlrobins (approved):
Thanks!

PR #229, opened 2024-12-18
Revise initial workspace -> sidecar websocket protocol

### LINKED ISSUE #228: Fix flakiness in test WebsocketEndpointTest.testWebsocketLifecycle()
Is flaky. Must deflake.

jlrobins:
## Summary of Changes

In an attempt to deflake tests, now require a workspace to be the agent that sends the first websocket message, a new WORKSPACE_HELLO message type, with a payload body that carries its workspace id. Previously, sidecar was the sender of the first message, and might have been sending it before the test suite workspace websocket client simulation was able to gets its message handler set up properly (as the theory goes about probable cause for test flakiness in seeming to "miss" the first message).
  * New connections are initially modeled as 'inactive', but kept in the single map.
  * Sessions migrate from inactive to active upon successful WORKSPACE_HELLO message receipt.
  * Sidecar uses the hello message body's workspace id to cross-reference against known workspace ids.
  * This negates the websocket route from needing a workspace_id query string parameter.

Having extension-side send the first message hopes to deflake test testWebsocketLifecycle, the theory being that without this, then sidecar-side would send the first message prior to when the test websocket client could establish its message handler. Now, with the client sending the first message, the test websocket client-side message handler will always be installed before the response WORKSPACE_COUNT_CHANGED gets received. Hopefully.

Many more tests written, all methods, > 90% of websocket-related lines. Getting coverage over the IOException handling lines remain.

* Known todo: periodic task to clean out too-long-in-purgatory websocket connections. Lower priority than un-flaking.
* Closes #228

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [x] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/main/java/io/confluent/idesidecar/websocket/messages/MessageType.java
```diff
@@ -15,16 +16,43 @@
...
+  /** Message sent by sidecar to a workspace when sidecar has noticed a websocket messaging
+   * error and is going to disconnect its end of the websocket. */
   PROTOCOL_ERROR,
 
   /** Placeholder for unknown-to-sidecar message types for messages intended to be
    * for extension -> extension messaging via sidecar. */
   UNKNOWN;
 
+  /**
+   * Static mapping of MessageType to the expected class of the body.
```
jlrobins, line 39: Lets deserializeMessage() perform a strong check.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
```diff
@@ -177,66 +136,39 @@ public void onOpen(Session session) throws IOException {
...
-      sendErrorAndCloseSession(
-          senderSession,
-          null,
-          "Odd! Received message from unregistered session %s. Closing session.",
-          senderSession.getId()
-      );
-      return;
-    }
+      // Expect to find them in purgatory, and that this is a HELLO_WORKSPACE message
+      if (!purgatorySessions.remove(senderSession)) {
```
rhauch, line 140: This line tries to remove the session from purgatory, but if it is successful don't we need to add it to the tracked sessions? It seems like this is done in `handleHelloMessage(...)`, but what happens if that fails? Maybe the removal from purgatory set should be done in the same place we're adding the session to the other map, so that they're always done together (without anything else between these two steps that might fail). Consider being defensive, and removing the session from purgatory _after_ it's added to the other map.

Maybe change this line to:
```suggestion
      if (!purgatorySessions.contains(senderSession)) {
```

Another way to simplify this is to not have separate collections for purgatory and validated workspace sessions, and instead have _one_ map that tracks both states and makes it easier to transition the state of any session. For example, maybe consider separating the logic into a separate `WebsocketSessions` class that encapsulates the map, the state management, and maybe even the `broadcast` logic that could stream over all but filter and skip any sessions that are in purgatory. That class could be an `@ApplicationScoped` bean and `@Injected` into the `WebsocketEndpoints` class, and that would also be a good spot to add in the future the scheduled method that purges sessions with expired TTL from the purgatory state. The public interface of this bean should hide all of the details and would only have the methods needed outside the class, such as getting the `WorkspaceSession` object for either a workspace ID or `Session` object (whichever is needed).

There are other benefits of this approach. 
1. Other classes/beans could broadcast events to sessions by using the `WebsocketSessions` bean, and then leave `WebsocketEndpoints` to be just that -- the endpoints and methods called by the Quarkus framework due to client calls.
2. The `WebsocketSessions` class and the logic we choose to encapsulate is much more easily unit tested, as it's more narrowly scoped than `WebsocketEndpoints`. 

Also, consider a `WorkspaceId` record type, so that the signatures of collections and methods are more strongly typed than with just `String`.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
```diff
@@ -265,16 +197,101 @@ public void onMessage(String messageString, Session senderSession) {
...
+   *
+   * The session is always removed from the purgatory set.
+   *
+   * @param session the session that sent the message.
+   * @param messageString the JSON string message.
+   */
+  private void handleHelloMessage(Session session, String messageString) {
+    // Regardless of if the message was good or bad, we remove the session from purgatory
+    // upon receipt of the first message.
+    purgatorySessions.remove(session);
```
rhauch, line 216: I'd recommend removing from purgatory later on, just before the session is added to the other map. Lots of things might go wrong here, and this method is not using try-with-finally blocks to ensure the session is either left as-is or both removed from the set and added to the map.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
```diff
@@ -265,16 +197,101 @@ public void onMessage(String messageString, Session senderSession) {
...
+            message.headers().id(),
+            "Workspace id %s already connected. Closing session.",
+            workspaceId
+        );
+        return;
+      }
+    }
+
+    // All good! Add the workspace to the sessions map and announce the count change.
+    sessions.put(session, new WorkspaceSession(workspaceId));
```
rhauch, line 269: What do you think about having the `WorkspaceSession` record hold a reference to the `Session` object? That way we can return a `WorkspaceSession` object and that also provides access to the `Session` object?

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
```diff
@@ -265,16 +197,101 @@ public void onMessage(String messageString, Session senderSession) {
...
   @OnError
   public void onError(Session session, Throwable throwable) throws IOException {
     // May or may not actually remove -- if had not yet been authorized, it won't be in the map.
     // (but will definitely not be in the map after this statement.)
     var existingSession = sessions.remove(session);
+    if (existingSession == null) {
+      // Must have been in purgatory and had an error before saying hello.
+      purgatorySessions.remove(session);
+    }
+
```
rhauch, line 289: Using a single collection would also improve this logic and guarantee it either happens or does not happen atomically. 

Sorry for pointing you toward a separate "purgatory" collection. :-(

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
```diff
@@ -385,4 +405,107 @@ static MessageHeaders validateHeadersForSidecarBroadcast(Message outboundMessage
...
+          sendErrorAndCloseSession(
+              session,
+              headers.id(),
+              "Expected %s message body, got %s. Closing session.",
+              expectedBodyType.getSimpleName(),
+              body.getClass().getSimpleName()
+          );
+          return null;
+        }
+      }
```
rhauch, line 465: Okay, this is very nice and useful in helping us catch programming errors where the message type header and the body object are of different types. Well done!

rhauch (commented):
Thanks for taking a stab at this. I do have some minor suggestions and one major one: encapsulate the tracked sessions in a new `@ApplicationScoped` bean class that has _one_ collection for all sessions, and track whether a session as bound to a workspace as a state. Let me know if you want to talk this over.

jlrobins:
Will have good revisions posted tonight for tomorrow morning, thanks.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
```diff
@@ -36,6 +43,14 @@ public class WebsocketEndpoint {
   @VisibleForTesting
   final Map<Session, WorkspaceSession> sessions = new ConcurrentHashMap<>();
 
+  /**
+   * Set of sessions which have connected, but not yet sent a valid hello message.
+   * The time that they connected is stashed as session.getUserProperties().get("connectedAt")
+   * so that they can be purged after a certain amount of time if they linger in this state.
+   */
+  @VisibleForTesting
+  final Set<Session> purgatorySessions = Collections.synchronizedSet(new HashSet<>());
```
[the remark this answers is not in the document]
jlrobins: Purgatory map no more, only the single concurrent map of sessions, be they pre- or post- hello, that state described by the `WorkspaceWebsocketSession::isActive()`.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
jlrobins: Going with your provided revisions of wholistic map of not-yet-hello / post-hello workspace sessions, but directly in the websocket route resource.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
jlrobins: Done.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
jlrobins: Is doing that way now.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
jlrobins: Done.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
```diff
@@ -70,17 +146,30 @@ public void broadcast(Message message) throws IOException, IllegalArgumentExcept
...
+      try {
+        f.get();
+      } catch (Exception e) {
+        Log.errorf("Failed to send broadcast message to workspace: %s", e.getMessage());
+
+        if (e instanceof InterruptedException) {
+          Thread.currentThread().interrupt();
+        }
+      }
+    }
```
rhauch, line 172: We don't want to block here. First, it might block forever and block the thread. Second, it's really an anti-pattern in Quarkus.

The proper way to do this is to return some sort of promise, which in Quarkus is `Uni` for single objects or `Multi` for multiple items. So the first step is to add a helper method to send a message asynchronously in `WebsocketWorkspaceSession`:
```
    /**
     * Send the text message with the given message ID, and return a {@link Uni} that returns
     * the message ID when the message is sent.
     *
     * @param messageText
     * @param messageId
     * @return the message ID when the message is sent.
     */
    public Uni<String> sendAsync(String messageText, String messageId) {
      Log.debugf(
          "Sending broadcast message %s to workspace pid %d",
          messageId,
          workspacePid.id()
      );

      var future = CompletableFuture.supplyAsync(() -> {
        session.getAsyncRemote().sendText(messageText, result -> {
          if (result.isOK()) {
            Log.debugf(
                "Successfully sent broadcast message %s to workspace pid %d",
                messageId,
                workspacePid.id()
            );
          } else {
            var cause = result.getException();
            Log.errorf("Failed to send broadcast message to workspace: %s", cause.getMessage());
            throw new CompletionException(cause);
          }
        });
        return messageId;
      });
      return Uni.createFrom().completionStage(future);
    }
```
Then we can have multiple forms of `broadcast`:
```
  /**
   * Broadcast a message originating from sidecar to all active connected workspaces.
   *
   * @param message the message to broadcast
   * @throws IOException              if there is an error serializing the message to JSON.
   * @throws IllegalArgumentException if the headers of the message are not valid for broadcasting.
   * */
  public Multi<String> broadcast(Message message) throws IOException, IllegalArgumentException {
    return broadcast(message, WorkspaceWebsocketSession::isActive, "active");
  }

  /**
   * Broadcast a message originating from sidecar to connected workspaces that match the supplied
   * filter.
   *
   * @param message           the message to broadcast
   * @param filter            the filter to apply to the workspace sessions
   * @param sessionsAdjective the adjective to describe the sessions in the log message.
   * @throws IOException              if there is an error serializing the message to JSON.
   * @throws IllegalArgumentException if the headers of the message are not valid for broadcasting.
   * */
  public Multi<String> broadcast(
      Message message,
      Predicate<WorkspaceWebsocketSession> filter,
      String sessionsAdjective
  ) throws IOException, IllegalArgumentException {
    var activeWorkspaceSessions = sessions.values().stream()
                                          .filter(filter)
                                          .toList();

    if (activeWorkspaceSessions.isEmpty()) {
      Log.debugf("No %s workspaces to broadcast message to.", sessionsAdjective);
      return Multi.createFrom().empty();
    }

    final MessageHeaders headers = validateHeadersForSidecarBroadcast(message);

    // Serialize the message to JSON
    String jsonMessage = mapper.writeValueAsString(message);
    Log.debugf(
        "Broadcasting %d char message, id %s to all workspaces",
        jsonMessage.length(),
        headers.id()
    );

    return broadcast(jsonMessage, headers.id(), filter, sessionsAdjective);
  }

  /**
   * Broadcast a message originating from sidecar to connected workspaces that match the supplied
   * filter.
   *
   * @param jsonMessage       the message to broadcast
   * @param messageId         the ID of the message (used for logging)
   * @param filter            the filter to apply to the workspace sessions
   * @param sessionsAdjective the adjective to describe the sessions in the log message.
   * @throws IOException              if there is an error serializing the message to JSON.
   * @throws IllegalArgumentException if the headers of the message are not valid for broadcasting.
   * */
  public Multi<String> broadcast(
      String jsonMessage,
      String messageId,
      Predicate<WorkspaceWebsocketSession> filter,
      String sessionsAdjective
  ) {
    // Compute the sessions that satisfy the filter
    var activeWorkspaceSessions = sessions.values().stream()
                                          .filter(filter)
                                          .toList();
    if (activeWorkspaceSessions.isEmpty()) {
      Log.debugf("No %s workspaces to broadcast message to.", sessionsAdjective);
      return Multi.createFrom().empty();
    }

    var promises = activeWorkspaceSessions
        .stream()
        .map(session -> session.sendAsync(jsonMessage, messageId))
        .toList();
    return Multi
        .createFrom()
        .iterable(promises)
        .onItem()
        .transformToMultiAndMerge(Uni::toMulti);
  }
```
If a caller doesn't care about waiting, they can ignore the `Multi` response. But if they do, they can do [various things](https://smallrye.io/smallrye-mutiny/latest/guides/imperative-to-reactive/), including [handling errors](https://smallrye.io/smallrye-mutiny/latest/tutorials/handling-failures/).

For example, a test might want to block until the broadcast is completed:
```
    endpoints.broadcast(...).collect().asList().await().atMost(Duration.ofSeconds(5));
```

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
```diff
@@ -174,124 +204,190 @@ public void onOpen(Session session) throws IOException {
...
-          "Proxying message %s from workspace %d to %d other workspace(s)",
-          headers.id(),
-          workspaceSession.processId(),
+          "Proxying message %s from workspace %s to %d other workspace(s)",
+          m.id(),
+          workspaceSession.workspacePid(),
           otherCount
       );
+
+      List<Future> futures = new ArrayList<>();
```
rhauch, line 257: See comment above about not blocking here, either.
```
    Log.debugf(
        "Received message %s of messageType %s from workspace %s",
        m.id(),
        m.messageType(),
        workspaceSession.workspacePid()
    );

    // At this time, all websocket messages received from workspaces are intended to
    // be proxied to all the other workspaces. Do so here.
    broadcast(
        messageString,
        m.id(),
        wws -> wws.session().isOpen() && wws.isActive() && wws.workspacePid() != workspaceSession.workspacePid(),
        "other"
    );
  }
```

rhauch (commented):
Thanks, @jlrobins. Some suggestions about not blocking in our production code.

rhauch:
See #234 for my suggestions in an easier-to-read format.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
```diff
@@ -98,72 +183,15 @@ public void broadcast(Message message) throws IOException, IllegalArgumentExcept
    */
   @OnOpen
   public void onOpen(Session session) throws IOException {
-    Log.info("New websocket session opened: " + session.getId());
+    Log.infof("New websocket session opened, not yet considered active: %s", session.getId());
+    Log.infof("websocketEndpoint: %s", this);
+    Log.infof("websocketEndpoint: sessions %s", System.identityHashCode(sessions));
```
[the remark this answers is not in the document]
jlrobins: they were removed in subsequent commit.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
```diff
@@ -70,17 +146,26 @@ public void broadcast(Message message) throws IOException, IllegalArgumentExcept
...
+    });
+
+    // Wait for all the futures to complete.
+    for (Future f : futures) {
+      try {
+        f.get();
+      } catch (Exception e) {
+        Log.errorf("Failed to send broadcast message to workspace: %s", e.getMessage());
+      }
+    }
```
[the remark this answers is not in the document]
jlrobins: Accepted your revisions from side branch, thank you.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
jlrobins: Accepted your side-branch, thank you.

src/main/java/io/confluent/idesidecar/websocket/resources/WebsocketEndpoint.java
```diff
@@ -5,36 +5,108 @@
...
+  }
+
   /**
-   * Map of active, authorized workspace sessions, keyed by the websocket session object.
+   * Map of all workspace sessions (valid or not), keyed by the websocket
+   * session id wrapped as SessionKey.
    */
-  @VisibleForTesting
-  final Map<Session, WorkspaceSession> sessions = new ConcurrentHashMap<>();
+  Map<SessionKey, WorkspaceWebsocketSession> sessions = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
jlrobins: Fixed.

rhauch (approved):
Thanks for gritting through this change, @jlrobins. Looks great, pending a green build.

rhauch (approved):
LGTM

PR #231, opened 2024-12-19
chore: Configure Sonarqube main branch for ide-sidecar [ci skip]

rohitsanj:
## Summary of Changes

Configure Sonarqube main branch to `main`, overrides default `master`.


flippingbits (approved):
Thanks, @rohitsanj! I don't think this change will take any effect, given that the SonarQube project has already been created, but let's give it a try.

PR #171, opened 2024-11-18
Implement topic config routes in Internal Kafka REST

rohitsanj:
Resolves #170 

## Summary of Changes

Implemented topic config list and batch alter topic configs endpoints in Internal Kafka REST. Added new integration tests for local and direct connections.

## Any additional details or context that should be provided?

The newly added routes may now be invoked from the public Kafka REST proxy as follows:

Use this fully-reproducible [test script](https://gist.github.com/rohitsanj/8f27be4676e463880dd64cb79de5efb2) to verify the changes in this branch. 

```shell
# List topic configs -- should see value as 'delete'
curl "http://localhost:26636/kafka/v3/clusters/${cluster_id}/topics/${topic_name}/configs" \
-H "x-connection-id: ${connection_id}" \
-H "Authorization: Bearer ${DTX_ACCESS_TOKEN}" | jq '.data[] | select(.name == "cleanup.policy")'

# Alter topic configs
curl "http://localhost:26636/kafka/v3/clusters/${cluster_id}/topics/${topic_name}/configs:alter" \
-H "Authorization: Bearer ${DTX_ACCESS_TOKEN}" \
-H "x-connection-id: ${connection_id}" \
-H "content-type: application/json" \
-d '{
  "data": [
    {"name": "cleanup.policy", "value": "compact", "operation": "set"}
  ]
}'

# List topic configs again -- should see value as 'compact'
curl "http://localhost:26636/kafka/v3/clusters/${cluster_id}/topics/${topic_name}/configs" \
-H "x-connection-id: ${connection_id}" \
-H "Authorization: Bearer ${DTX_ACCESS_TOKEN}" | jq '.data[] | select(.name == "cleanup.policy")'

# Now delete the override
curl "http://localhost:26636/kafka/v3/clusters/${cluster_id}/topics/${topic_name}/configs:alter" \
-H "Authorization: Bearer ${DTX_ACCESS_TOKEN}" \
-H "x-connection-id: ${connection_id}" \
-H "content-type: application/json" \
-d '{
  "data": [
    {"name": "cleanup.policy", "operation": "delete"}
  ]
}'


# List topic configs again -- should see value as 'delete'
curl "http://localhost:26636/kafka/v3/clusters/${cluster_id}/topics/${topic_name}/configs" \
-H "x-connection-id: ${connection_id}" \
-H "Authorization: Bearer ${DTX_ACCESS_TOKEN}" | jq '.data[] | select(.name == "cleanup.policy")'
```

[...]

PR #250, opened 2024-12-20
chore: Remove dead code and scripts

rohitsanj:
## Summary of Changes

- We're not using any of the `build-tools` scripts, removed those.
- Also removed some unused classes - `TopicConfig` and `AbstractConfig`, turns out I'd forgotten to remove these before merging #171.


src/main/java/io/confluent/idesidecar/restapi/kafkarest/TopicConfig.java
jlrobins, line 1: Am pretty sure we're making topic config calls -- is that being transparently proxied and doesn't need sidecar-specific classes and/or those aren't this class?

rohitsanj: Those aren't this class, yep. I'd added this thinking I'd use it but never did and it skipped in.

jlrobins (approved):

PR #266, opened 2025-01-03
Integrate with Scaffolding Service

### LINKED ISSUE #245: Integrate with Scaffolding Service
We want to replace the sidecar's scaffolding functionality, managed in the package `io.confluent.idesidecar.scaffolding`, with a proxy for the CCloud Scaffolding Service.

This will allow us to delete the package `io.confluent.idesidecar.scaffolding` and the folder `src/main/resources/static/templates`. We can remove the Make targets [zip-templates and zip-templates-windows](https://github.com/confluentinc/ide-sidecar/blob/main/Makefile#L32-L40), and delete the references to them from the `pom.xml` ([here](https://github.com/confluentinc/ide-sidecar/blob/main/pom.xml#L453-L467) and [here](https://github.com/confluentinc/ide-sidecar/blob/main/pom.xml#L480-504)).

rohitsanj:
## Summary of Changes

- Delegate list, get and apply template functionality to the Confluent Cloud Scaffolding Service.
- Added unit tests with mocked responses from the Scaffolding Service.

Resolves #245 

Upon click testing the executable from this branch in the vscode extension, I noticed a slight difference in the apply template behavior for some templates that enforce a strict `pattern` on input options. We'll need to augment the Scaffold HTML form in the vscode extension to set the pattern against input fields (cc @noeldevelops, I can take up this change with your guidance).

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [x] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/main/java/io/confluent/idesidecar/restapi/exceptions/ScaffoldingException.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Exception thrown when there is an error in the template generation process.
+ */
+@RegisterForReflection
+public class ScaffoldingException extends RuntimeException {
+
+  private final Failure failure;
+
+  public ScaffoldingException(String message, Failure failure, Throwable t) {
```
Cerchie, line 14: this is unused, should it be kept?

src/generated/resources/openapi.yaml
```diff
@@ -1307,25 +1334,25 @@ components:
...
-          enum:
-          - UP
-          - DOWN
+        name:
           type: string
         data:
           type: object
           nullable: true
-        name:
+        status:
```
rohitsanj, line 1354: Gah.. I'll remove this in a bit.

src/test/java/io/confluent/idesidecar/restapi/resources/TemplateResourceIT.java
```diff
@@ -1,118 +0,0 @@
...
-  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
-  private static final SecureRandom SECURE_RANDOM = new SecureRandom();
-  private static final int RANDOM_STRING_LENGTH = 25;
-
-  @Test
-  void shouldBeAbleToApplyAllTemplates() throws JsonProcessingException {
-    // Get all templates
-    var responseBody = given()
-        .when()
-        .get("/gateway/v1/templates")
```
Cerchie, line 35: this is the same url now yeah? would we want to test this functionality still at the proxy level?

Cerchie (commented):
Screenshot 2025-01-06 at 12 14 00 PM
Nice!

src/test/java/io/confluent/idesidecar/restapi/resources/TemplateResourceIT.java
rohitsanj: Hm, I don't think we _can_ test the functionality at proxy level using integration tests unless we hit stag/devel CCloud service and I'm not sure we want to do that. Testing that payloads are sent and received properly from the service should be sufficient and this is tested in `TemplateResourceTest`.

Makefile
```diff
@@ -29,16 +29,6 @@ test-native:
 clean:
 	$(MAKE) mvn-clean
 
-.PHONY: zip-templates
-zip-templates:
-	(cd src/main/resources/static && zip -r templates.zip templates/)
```
flippingbits, line 34: Can we also remove the folder `src/main/resources/static/templates` from the repository? This would reduce the size of the native executable.

src/main/java/io/confluent/idesidecar/restapi/resources/TemplateResource.java
```diff
@@ -29,30 +46,65 @@
...
-    var templateManifests = templateRegistryService.listTemplates();
-    var templateList = templateManifests.stream()
-        .map(Template::new)
-        .sorted(Comparator.comparing(Template::id))
-        .toList();
-    return new TemplateList(templateList);
+  public Uni<TemplateList> list() {
+    return uniStage(() -> scaffoldingProxyProcessor.process(LIST_TEMPLATES_PROXY_CONTEXT).toCompletionStage())
+        .chain(context -> parseJson(
+            context.getProxyResponseBody().toString(), ScaffoldV1TemplateList.class))
```
flippingbits, line 88: Did you decide to parse the CCloud service's JSON response and return a `Uni<TemplateList>` here (and in `get()`) so that we can auto-generate the OpenAPI spec?

flippingbits (commented):
Excellent work, @rohitsanj! I have two minor questions.

Makefile
rohitsanj: Yes, I figured I'd do the cleanup in a follow up PR, along with removing the scaffolding package as well.

src/main/java/io/confluent/idesidecar/restapi/resources/TemplateResource.java
rohitsanj: Yeah, it's so that there's no migration needed from the VS Code calling side.

flippingbits (approved):
Thanks, @rohitsanj! LGTM from reading, assuming you'll remove the static templates and scaffolding package in a follow-up PR shortly.

src/generated/resources/openapi.yaml
rohitsanj: I'll revert this in the cleanup PR, I don't feel like waiting another 20 mins just for this.

src/main/java/io/confluent/idesidecar/restapi/exceptions/ScaffoldingException.java
rohitsanj: It should be fine to keep this.

PR #270, opened 2025-01-03
Fix connection status update by ignoring stale refreshes

### LINKED ISSUE #267: Ensure connection updates do not race with scheduled connection status refreshes
It was observed while testing the edit connections PR https://github.com/confluentinc/vscode/pull/798 that going from invalid -> valid connection credentials or other way round returns stale connections status (of `FAILED` or `SUCCESS` respectively), and only eventually gets to the right status on a refresh. We believe this is due to the scheduled connection refresh racing with the connection update. These operations need to be synchronized.

rohitsanj:
## Summary of Changes

Fixes #267 

- Store the `lastUpdated` time for the connection status as an atomic reference. This is used to prevent stale connection status refreshes from going through.

**Before:**

1. [Sidecar] Refresh connection status task 1 kicks off using ConnectionSpec A (say, with valid credentials)
2. [VS Code] User updates connection with invalid Kafka credentials
3. [Sidecar] Update connection spec (now A') processed and Connection status set to `ATTEMPTING` by sidecar. 
4. [VS Code] Sees `ATTEMPTING` status and starts polling the connection status until `SUCCESS` or `FAILED`.
5. [Sidecar] Refresh connection status task 1 completes and updates connection status to `SUCCESS`.
6. [VS Code] Sees `SUCCESS` in subsequent poll and reports to user. -- **This is the wrong state!** ❌ 
7. [Sidecar] Refresh connection status task 2 kicks off using Connection Spec A' and sets connection status to `FAILED`. -- This is the correct state, but VS Code will not pick it up automatically.

**After:**

1. [Sidecar] Refresh connection status task 1 kicks off using ConnectionSpec A (say, with valid credentials). **Store current time here**
2. [VS Code] User updates connection with invalid Kafka credentials
3. [Sidecar] Update connection spec (now A') processed and Connection status set to `ATTEMPTING` by sidecar. **Set last modified time to now**
4. [VS Code] Sees `ATTEMPTING` status and starts polling the connection status until `SUCCESS` or `FAILED`.
5. [Sidecar] Refresh connection status task 1 completes. Sees that last modified time is _later_ than when we started this refresh, so skip the connection status update.
4. [VS Code] Still sees `ATTEMPTING` status on continuous polling.
7. [Sidecar] Refresh connection status task 2 kicks off using Connection Spec A' and sets connection status to `FAILED`.
9. [VS Code] Sees `FAILED` status and reports to user. -- **This is correct** ✅ 

Click tested with @shouples 

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionState.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionState.java
@@ -111,7 +111,7 @@ public abstract class ConnectionState {
    *
    * <p>This method always calls {@link #doRefreshStatus()} and then on success updates
    * the {@link #getStatus() cached connection status}. If we see that the status has been updated
-   * since the refresh was started, we ignore the update and refresh the status again.
+   * since the refresh was started, we ignore the update.
    *
    * @return the future that will complete with the updated connection status
    * @see #getConnectionStatus()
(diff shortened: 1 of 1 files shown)


shouples (approved):
LGTM - paired and click-tested a native executable against the extension and the issues we saw with stale `ConnectedState`s are no more! 👏

src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionState.java
```diff
@@ -99,24 +110,54 @@ protected ConnectionStatus getInitialStatus() {
...
+            lastUpdatedInstant,
+            beforeStartingRefresh
+        );
+      }
+    });
   }
 
   private void updateStatus(ConnectionStatus original, ConnectionStatus updated) {
     // update the cached status
     this.cachedStatus.set(updated);
```
jlrobins, line 153: Updating two 'atomic' references serially w/o an overarching lock still smells race-condition-y. An observer could observe the updated cachedStatus w/o the updated lastUpdated, couldn't they?

rohitsanj: Hm, I'm not sure if there was ever a "race" at all, actually. This change makes it such that refreshes started before an update don't go through. 

Currently, the only observer reading the `lastUpdated` atomic value is the refresh task. And there is always only a single instance of this per connection type. Hence, there is no possibility for multiple observers to race and read stale `lastUpdated` values.

rohitsanj: Added some details regarding the before/after state to the PR description.

PR #289, opened 2025-01-09
Return status code 400 when updating unauthed CCloud connection

### LINKED ISSUE #287: When updating a CCloud connection, return the response status code 400 in case of errors
When clients update a CCloud connection, the sidecar checks if it can, for instance, [authenticate with the configured CCloud organization](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionStateManager.java#L235). There might be cases in which the sidecar returns a `401` status if something goes wrong when interacting with the CCloud API although a `400` status might be the more appropriate error and is also [documented](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionStateManager.java#L261).

We can fix this, i.e., return a `400` instead of a `401` status in case of issues, by updating the [exception mapper for the CCloudAuthenticationFailedException](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/exceptions/ExceptionMappers.java#L113) or by introducing a new exception that is thrown if this validation fails (and setting up the related exception mapper correctly).

flippingbits:
Return the HTTP response status code 400 not 401 when attempting to update an unauthenticated CCloud connection.

Fixes #287

## Pull request checklist

- Tests:
    - [ ] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionStateManager.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/connections/ConnectionStateManager.java
@@ -265,21 +265,47 @@ public class ConnectionStateManager {
     var connectionState = (CCloudConnectionState) connectionStates.get(id);
     CCloudOAuthContext authContext = connectionState.getOauthContext();
 
-    return Uni.createFrom()
-        .completionStage(authContext
-            .refreshIgnoreFailures(organizationId).toCompletionStage())
-        .onFailure(CCloudAuthenticationFailedException.class).recoverWithUni(
+    return Uni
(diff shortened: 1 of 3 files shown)


rohitsanj (approved):
LGTM, thanks!

rohitsanj (approved):

PR #290, opened 2025-01-09
Fix OAuthCredentials (and connecting to SR via direct connections)

### LINKED ISSUE #126: Add OAuth 2.0 authN to direct connections
Continuation of #124 and #125 that added basic and mTLS authN support to direct connections. 

We want to add OAuth support to direct connections (similar to basic creds) via the `kafka_cluster` and `schema_registry` definitions, and use this to connect to downstream resources (e.g., Kafka REST API, SR API endpoints, and Message Consumer API). All secrets in credentials will be write-only: the Connection REST API endpoints must always mask secrets. 

We'll track adding support for other credential types in separate issues:
* basic authN (#124)
* mTLS authN (#125)
* OAuth authN (this issue)
* Kerberos authN, or SSL/GSSAPI (#127)

rohitsanj:
## Summary of Changes

Closes #126 

Changes related to fixing OAuth credentials:
- Adds a new credential option for `logicalClusterId` -- this is used when connecting to CCloud Kafka/SR over OAuth. See this [docs page](https://docs.confluent.io/cloud/current/security/authenticate/workload-identities/identity-providers/oauth/overview.html).
- Rename `identityPool` to `identityPoolId` for clarity
- Fix JAAS config construction by adding a semicolon at the end.

Once I fixed the above, I noticed that Kafka requests were going through fine but the SR requests were all failing with a 401. After some investigation, I found that we need to implement the `httpAuthenticationHeaders` method for each `Credentials` type. This is then used as part of the SR proxy implementation.

Implementing `httpAuthenticationHeaders` is simple enough for `BasicCredentials` (or `ApiKeyAndSecret`) since it's only a matter of base64 encoding the `username:password` string, however, for `OAuthCredentials`, this is likely much more complex given the breadth of configs available. 

The remaining changes in the PR:

- The SR proxy implementation now uses the cached SchemaRegistryClient instance to issue the request instead of constructing the full HTTP request ourselves.

I tested the OAuth credentials against Confluent Cloud by connecting to a Kafka cluster. Followed [this guide](https://docs.confluent.io/cloud/current/security/authenticate/workload-identities/identity-providers/oauth/identity-providers.html) to setup Okta as an OAuth provider.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [x] Updated existing
    - [x] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/main/java/io/confluent/idesidecar/restapi/clients/ClientConfigurator.java
```diff
@@ -318,18 +318,10 @@ public static Map<String, Object> getSchemaRegistryClientConfig(
...
-        .map(CCloud.LsrcId.class::cast)
-        .map(CCloud.LsrcId::id)
-        .orElse(null);
-
     // Add any properties for SR credentials (if defined)
     var options = connection
         .getSchemaRegistryOptions()
-        .withRedact(redact)
-        .withLogicalClusterId(logicalId);
+        .withRedact(redact);
```
rohitsanj, line 324: The Confluent Cloud logical cluster id for a schema registry will need to be provided by the user as part of the credentials config in case of OAuth credentials. We cannot infer the "logical" cluster id from the Schema Registry URI which may or may not contain the cluster id in its first part. Moreover, a given physical cluster id may host multiple logical cluster ids.

src/main/java/io/confluent/idesidecar/restapi/connections/DirectConnectionState.java
```diff
@@ -241,8 +224,8 @@ protected Future<KafkaClusterStatus> getKafkaConnectionStatus() {
   protected Future<SchemaRegistryStatus> getSchemaRegistryConnectionStatus() {
     return withSchemaRegistryClient(srClient -> {
       // There is a configuration, so validate the connection by creating a SchemaRegistryClient
-      // and getting all subjects.
-      srClient.getAllSubjects();
+      // and getting all schema types
+      srClient.getSchemaTypes();
```
rohitsanj, line 228: `getSchemaTypes` is much less expensive than a `getAllSubjects` call. This is kind of unrelated to the PR's changes.

src/main/java/io/confluent/idesidecar/restapi/credentials/OAuthCredentials.java
```diff
@@ -55,11 +57,19 @@ public record OAuthCredentials(
   @Null
   Integer connectTimeoutMillis,
 
+  @Schema(description = "Additional property that can be added in the request header to identify "
+      + "the logical cluster ID to connect to. For example, this may be a Confluent Cloud " +
+      "Kafka or Schema Registry cluster ID.")
+  @JsonProperty(value = "logical_cluster_id")
+  @Null
+  String logicalClusterId,
```
flippingbits, line 65: Do you know if the `logicalClusterId` is needed for clusters other than CCloud ones? If not, what about modeling this in a bit more generic way so that we can maybe support other `extension_`s as well without having to introduce new fields to the `OAuthCredentials`?

flippingbits (commented):
Great work, @rohitsanj! I have a few minor comments. Otherwise, your PR looks good to me!

src/main/java/io/confluent/idesidecar/restapi/credentials/OAuthCredentials.java
rohitsanj: Hm, yeah it is definitely CCloud specific, but so is the other `identityPoolId` field. 

> If not, what about modeling this in a bit more generic way so that we can maybe support other extension_s as well without having to introduce new fields to the OAuthCredentials?

Hm, would it be fine to revisit this at a later point? For now I imagine that we will have some client-side logic that will only surface these fields for direct connections to Confluent Cloud.

flippingbits: Thanks for the clarification! From my POV, it's fine to add these two fields now because we probably won't need to support any other _custom_ JAAS configs in the near/mid term.

> Hm, would it be fine to revisit this at a later point?

I don't think we can remove them at a later point as they'll be part of the API contract and Direct Connection form. Removing them would have an impact on existing Direction Connections of users and be a breaking change.

> For now I imagine that we will have some client-side logic that will only surface these fields for direct connections to Confluent Cloud.

Maybe make it clear that these fields are only relevant for Direct Connections that target CCloud clusters by, for instance, adding the prefix `ccloud_` to their names?

rohitsanj: > I don't think we can remove them at a later point as they'll be part of the API contract and Direct Connection form. Removing them would have an impact on existing Direction Connections of users and be a breaking change.

Makes sense. Thanks for mentioning.

> adding the prefix ccloud_ to their names?

Prefixing `ccloud_` sounds good. :+1:

PR #293, opened 2025-01-13
Introduce new CCloud config `ide_auth_callback_uri`

### LINKED ISSUE #288: Make VSCode Extension URI a property of the Connection spec's CCloud config
Upon completion of the authentication with Confluent Cloud, the sidecar redirects the user to the VS Code extension so that they can immediately start interacting with their Confluent Cloud resources in the extension. At the moment, the redirect URI is configurable via the application configuration `ide-sidecar.connections.ccloud.oauth.vscode-extension-uri` and is set to `vscode://confluentinc.vscode-confluent/authCallback` by default.

We want to support using the same sidecar instance with multiple instances of "VS Code distributions", e.g., the regular VS Code, VS Code Insiders, Cursor, etc. Each VS Code distribution brings its own URI schema, making it mandatory to use different redirect URIs for different CCloud connections. As a consequence, we want to allow the extension to set the VSCode Extension URI via the [ccloud_config](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java#L234) in the [Connection spec](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java#L47).

flippingbits:
Introduce the new CCloud config `ide_auth_callback_uri` to allow CCloud connections to specify their own custom URI that users will be redirected to after successfully completing the authentication flow with Confluent Cloud.

This replaces/extends the global redirect URI, which is defined via the application config `ide-sidecar.connections.ccloud.oauth.vscode-extension-uri` at sidecar startup time, and is needed to support multiple VS Code distributions/flavors, each having its own custom URI scheme, using the same sidecar instance.

If a CCloud connection does not define the `ide_auth_callback_uri`, the sidecar will continue using `ide-sidecar.connections.ccloud.oauth.vscode-extension-uri` as the redirect URI.

Fixes #288

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java
```diff
@@ -236,9 +235,16 @@ public record CCloudConfig(
...
+      @JsonProperty(value = "organization_id")
       @Size(min = 36, max = 36)
-      String organizationId
+      String organizationId,
+      @Schema(
+          description = "The URI that users will be redirected to after successfully completing "
+          + "the authentication flow with Confluent Cloud."
+      )
+      @JsonProperty(value = "ide_auth_callback_uri")
+      @Size(min = 10, max = 200)
```
flippingbits, line 246: I feel these are reasonable defaults for validating the `ide_auth_callback_uri`. It should have at least `len("vscode://") + 1` characters and is not expected to be longer than 200 characters.

src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java
```diff
@@ -236,9 +235,16 @@ public record CCloudConfig(
           description = "The identifier of the CCloud organization to use. "
                         + "The user's default organization is used when absent."
       )
-      @JsonProperty(value = "organization_id", required = true)
```
flippingbits, line 239: Since the CCloudConfig now holds multiple properties, I think it doesn't make sense to keep requiring the CCloud org to be set, as we'll use the user's default org in the absence of this property per the docs.

src/main/java/io/confluent/idesidecar/restapi/resources/OAuthCallbackResource.java
```diff
@@ -109,15 +114,40 @@ public Uni<String> callback(@QueryParam("code") String authorizationCode,
...
+  private Future<String> renderFailure(Throwable error, String redirectUri) {
     Log.error(error);
     var errorMessage = error instanceof SSLHandshakeException
         ? TLS_HANDSHAKE_ERROR_MESSAGE
         : error.getMessage();
     return Future.succeededFuture(
         callbackFailure
             .data("error", errorMessage)
-            .data("vscode_redirect_uri", CCLOUD_OAUTH_VSCODE_EXTENSION_URI)
+            .data("vscode_redirect_uri", Objects.requireNonNullElse(redirectUri, ""))
```
flippingbits, line 150: The callback page might be called with a non-CCloud connection or an invalid `state` param, in which case we don't have access to the connection-defined redirect URI.

@shouples In this case, do you want the sidecar to refrain from redirecting the user or do you want the sidecar to redirect the user to the redirect URI specified via the application config `ide-sidecar.connections.ccloud.oauth.vscode-extension-uri`?

src/main/java/io/confluent/idesidecar/restapi/resources/OAuthCallbackResource.java
```diff
@@ -109,15 +114,40 @@ public Uni<String> callback(@QueryParam("code") String authorizationCode,
...
+    var connectionSpec = connection.getSpec();
+
+    if (connectionSpec != null) {
+      var ccloudConfig = connectionSpec.ccloudConfig();
+      if (ccloudConfig != null && ccloudConfig.ideAuthCallbackUri() != null) {
+        return ccloudConfig.ideAuthCallbackUri();
+      }
+    }
+
+    return CCLOUD_OAUTH_VSCODE_EXTENSION_URI;
```
flippingbits, line 135: If the CCloud connection doesn't specify the `ide_auth_callback_uri`, we use the redirect URI set via the application config `ide-sidecar.connections.ccloud.oauth.vscode-extension-uri` by default.

@shouples I'd like to discuss this with you.

src/generated/resources/openapi.json
```diff
@@ -767,14 +767,19 @@
...
         "properties" : {
           "organization_id" : {
             "description" : "The identifier of the CCloud organization to use. The user's default organization is used when absent.",
             "maxLength" : 36,
             "minLength" : 36,
             "type" : "string"
+          },
+          "ide_auth_callback_uri" : {
+            "description" : "The URI that users will be redirected to after successfully completing the authentication flow with Confluent Cloud.",
+            "maxLength" : 200,
```
Cerchie, line 780: curious about why we chose 200? Thinking about 3.2.1 General Syntax on [this page ](http://www.faqs.org/rfcs/rfc2616.html)

src/main/java/io/confluent/idesidecar/restapi/models/ConnectionSpec.java
Cerchie: I see... I think I still have the question about why 200 in particular

shouples (dismissed):
LGTM and works well from the VS Code extension perspective! Tested with a native executable against https://github.com/confluentinc/vscode/pull/915 in VS Code Insiders, and the behavior was as expected:
image

Good stuff! 🎉

src/main/java/io/confluent/idesidecar/restapi/resources/OAuthCallbackResource.java
shouples: To close out this comment based on our offline discussion yesterday: we decided it made sense to not redirect the user in this case, since it could become confusing if the connection initially came from one product/distribution but something happened to cause the `state` param to become invalid and then the user would possibly be prompted with a different product/distribution redirect.

src/main/resources/templates/callback.html
```diff
@@ -25,7 +25,10 @@
   <script type="text/javascript">
     // Redirect to the VS Code extension's URI when the page is loaded
     window.onload = function() {
-        window.location.href = "{vscode_redirect_uri}?success=true";
+        const redirectUri = "{vscode_redirect_uri}";
+        if (redirectUri.length > 0) {
```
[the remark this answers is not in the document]
flippingbits: This code checks for the [empty string that's passed if no redirect should be performed](https://github.com/confluentinc/ide-sidecar/blob/ccloud-spec-vscode-extension-uri/src/main/java/io/confluent/idesidecar/restapi/resources/OAuthCallbackResource.java#L150) in particular. I'd keep it as it is because the redirect URI can be filled through a global config, too.

src/generated/resources/openapi.json
flippingbits: I chose it more or less randomly. All redirect URIs that I can think of have less than 200 characters. At the same time, this validation prevents passing very long (invalid?) strings via the `ide_auth_callback_uri`.

We can (slightly) decrease the `maxLength` constraint to , for instance, 100 or 150 characters; LMK if you want me to do so.

flippingbits:
@Cerchie @shouples Thanks a lot for your reviews and click testing!

I had to resolve a merge conflict. Could you take another look when you have time?

@Cerchie I got back to you in the threads. Let me know if you want me to update the `maxLength` constraint for the `ide_auth_callback_uri`.

shouples (approved):

src/generated/resources/openapi.json
Cerchie: I think 200 is fine

src/main/resources/templates/callback.html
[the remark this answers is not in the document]
Cerchie: ah ok

Cerchie (approved):

PR #298, opened 2025-01-15
Remove fragile timestamp from connection error states which causes sidecar to emit websocket messages every 5s.

jlrobins:
## Summary of Changes

- Drop the Instant field stamping when a connection was observed to have auth errors. When, say, a direct connection had an unhappy schema registry, the every 5s polling would create a new timestamp in the connection error, ultimately causing sidecar to emit the updated-only-with-timestamp-change Connection as a CONNECTION_CHANGE websocket event to all the connected workspaces, adding work and noise and would need additional code to dejitter vscode-extension side. If extension side needed to care about when the error started, it will use the timestamp at which it received the single updated Connection that edged to unhappy as a "good enough" timestamp (and better than one that moves forward every 5s while in this general condition).
- Needed to more simply solve https://github.com/confluentinc/vscode/issues/908

- I have no suggestion as to the other changes made to the openapi documents. I wonder if prior PRs didn't commit their pending openapi document changes?

## Pull request checklist

- Tests:
    - [ ] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/generated/resources/openapi.json
+++ b/src/generated/resources/openapi.json
@@ -227,7 +227,7 @@
             }
           },
           "401" : {
-            "description" : "Could not authenticate with updated connection configuration",
+            "description" : "Could not authenticate",
             "content" : {
               "application/json" : {
                 "schema" : {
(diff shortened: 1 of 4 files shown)


src/generated/resources/openapi.json
```diff
@@ -227,7 +227,7 @@
             }
           },
           "401" : {
-            "description" : "Could not authenticate with updated connection configuration",
```
jlrobins, line 230: No idea about this change or the subsequent patch method block.

src/generated/resources/openapi.json
```diff
@@ -607,9 +669,6 @@
       "AuthError" : {
         "type" : "object",
         "properties" : {
-          "created_at" : {
```
jlrobins, line 610: This is the expected removal matching the codebase change.

src/main/java/io/confluent/idesidecar/restapi/auth/AuthErrors.java
```diff
@@ -79,7 +79,6 @@ public AuthErrors withoutTokenRefresh() {
   }
 
   public record AuthError(
-      @JsonProperty(value = "created_at") Instant createdAt,
```
jlrobins, line 82: The meat of the PR. Remove the createdAt field causing connections with "auth errors" for their kafka or schema registry to jitter over websocket every 5s check pass.

noeldevelops (approved):
PATCH endpoint came in yesterday to sidecar...maybe extra changes got generated as a result?

PR #301, opened 2025-01-16
Scram integration tests

flippingbits:
## Summary of Changes

This change adds a listener for SASL/SCRAM to the `CPServerContainer` so that we can verify the correctness of the `ScramCredentials` using the CP integration tests. For now, we run only the `TopicV3Suite` against the SASL/SCRAM listener but can add more test suites when needed.

The build is expected to fail right now because the build of the [base branch](https://github.com/confluentinc/ide-sidecar/tree/cerchie/scram) fails, too.

## Pull request checklist

- Tests:
    - [x] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/test/java/io/confluent/idesidecar/restapi/util/CPDemoTestEnvironment.java
+++ b/src/test/java/io/confluent/idesidecar/restapi/util/CPDemoTestEnvironment.java
@@ -4,7 +4,7 @@ import static org.junit.jupiter.api.Assertions.fail;
 
 import com.github.dockerjava.api.model.Container;
 import io.confluent.idesidecar.restapi.credentials.*;
-import io.confluent.idesidecar.restapi.credentials.Credentials.HashAlgorithm;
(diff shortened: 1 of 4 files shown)


src/test/java/io/confluent/idesidecar/restapi/util/CPDemoTestEnvironment.java
```diff
@@ -149,7 +157,7 @@ private void setMinISR() {
     try {
       kafka1.execInContainer(
           "kafka-configs",
-          "--bootstrap-server", "kafka1:12091",
+          "--bootstrap-server", "kafka1:12093",
```
flippingbits, line 160: I think `kafka1:12091` didn't work before inside the Docker network because this port is assigned to the `CLEAR` listener that is using `localhost` as the advertised listener.

PR #314, opened 2025-01-30
fix: Use `ProxyHTTPClient` when interacting with CCloud SR clusters

flippingbits:
This change makes sure we use the `ProxyHTTPClient` when proxying requests to CCloud SR endpoints, like `/schemas`, so that we take user-provided SSL certs and other custom HTTP configs passed via the Preferences API into account.

In follow-up work, we should build integration tests which verify that all CCloud proxies correctly pick up custom SSL certs and settings if configured. I'll capture this in a separate GH issue.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/proxy/ClusterProxyRequestProcessor.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/proxy/ClusterProxyRequestProcessor.java
@@ -50,10 +50,13 @@ public class ClusterProxyRequestProcessor extends
       // If Kafka, send the request to the Kafka REST server
       // using a generic HTTP client
       case KAFKA -> proxyHttpClient.send(context);
-      // If Schema Registry, use the cached SchemaRegistryClient instance
-      // to send the request to the Schema Registry server. Constructing the full
-      // HTTP request ourselves is error-prone and unnecessary.
-      case SCHEMA_REGISTRY -> processSchemaRegistry(context);
+      case SCHEMA_REGISTRY ->
+          switch (context.getConnectionState().getType()) {
+            // For CCloud connections, use Proxy HTTP client
+            case CCLOUD -> proxyHttpClient.send(context);
+            // For all other connections, use SR client
+            case DIRECT, LOCAL, PLATFORM -> processSchemaRegistry(context);
+          };
     };
 
     return clusterOp.compose(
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/idesidecar/restapi/proxy/ClusterProxyRequestProcessor.java
```diff
@@ -50,10 +50,13 @@ public Future<ClusterProxyContext> process(ClusterProxyContext context) {
...
-      // If Schema Registry, use the cached SchemaRegistryClient instance
-      // to send the request to the Schema Registry server. Constructing the full
-      // HTTP request ourselves is error-prone and unnecessary.
-      case SCHEMA_REGISTRY -> processSchemaRegistry(context);
+      case SCHEMA_REGISTRY ->
+          switch (context.getConnectionState().getType()) {
+            // For CCloud connections, use Proxy HTTP client
+            case CCLOUD -> proxyHttpClient.send(context);
+            // For all other connections, use SR client
+            case DIRECT, LOCAL, PLATFORM -> processSchemaRegistry(context);
```
jlrobins, line 58: Remind me why users won't possibly have similar SSL issues for DIRECT or PLATFORM schema registries? Because the direct connections will have already been SSL tuned accordingly?

flippingbits: Yes, it only applied to CCloud.

For the CCloud connection, we allow users to configure custom SSL certs via the extension settings. They're picked up by an [application-wide singleton instance of the WebClient](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/util/WebClientFactory.java#L75). We make sure that [updating the extension settings will update the WebClient's configuration](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/util/WebClientFactory.java#L89).

Direct/Platform connections rely on connection-specific SSL settings, which were correctly picked up.

shouples (approved):
Confirmed it works well when click-testing the native executable with the extension, and it also appears to fix user issues -- thanks for the quick work on this one @flippingbits!

Issue #81, opened 2024-10-14
Add RBAC/authz endpoint for CCloud resources


shouples:
CCloud uses https://confluent.cloud/api/metadata/security/v2alpha1/authorize and sends the following kind of `PUT` body structure:
```json
{
  "userPrincipal": "User:u-abc123",
  "actions": [
    {
      "resourceType": "Environment",
      "resourceName": "environment",
      "operation": "Alter",
      "scope": {
        "clusters": {},
        "path": [
          "organization=<orgId>",
          "environment=env-xyz987"
        ]
      }
    },
    {
      "resourceType": "Environment",
      "resourceName": "environment",
      "operation": "CreateSRCluster",
      "scope": {
        "clusters": {},
        "path": [
          "organization=<orgId>",
          "environment=env-xyz987"
        ]
      }
    }
  ]
}
```
And returns a basic `ALLOWED` (maybe sometimes `SUCCESS`?) or `DENIED` value for each action.
<img width="1802" alt="image" src="https://github.com/user-attachments/assets/e109c854-6025-4767-a9f0-211ad108c182">

From the VS Code extension perspective, this (or something similar, if available) will be required to know a user's authorized operations for various resources. Currently, the extension is only able to check accesses from below the (Kafka, maybe SR) cluster level (i.e. `authorized_operations` from a get/list topics response).

Cerchie:
Proxy to Kafka REST endpoint to return whether user has permissions to create topics, create/produce messages. SDS for CC resources. UKNOWN|SUCCESS|DENIED enum required for optimal FE.

PR #319, opened 2025-02-03
Cerchie/rbac

Cerchie:
adding RBAC capabilities. 

to test manually, run `make quarkus-dev` and  then: 

```
curl -H "Content-Type:application/json" -d'{"id": "6", "name": "1", "type": "CCLOUD"}' http://localhost:26636/gateway/v1/connections | jq
```
login via the sign in uri,

and then:

```
❯ curl -X PUT \
-H "Content-Type: application/json" \
-H "X-Connection-Id: 6" \
-H "Accept: application/json" \
-H "X-Debug: true" \
-d '{"userPrincipal":"User:YOUR_USER_ID","actions":[{"resourceType":"Organization","resourceName":"organization","operation":"CreateEnvironment","scope":{"clusters":{},"path":["organization=YOUR_ORG_ID"]}}]}' \
"http://localhost:26636/api/metadata/security/v2alpha1/authorize"                                                                                                                                                                     
```

should return
```
["ALLOWED"]%     
```


src/main/java/io/confluent/idesidecar/restapi/resources/RestProxyResource.java
```diff
@@ -115,5 +190,14 @@ private ClusterProxyContext createSRClusterContext(RoutingContext routingContext
         ClusterType.SCHEMA_REGISTRY
     );
   }
-
+  private ProxyContext createRBACProxyContext(RoutingContext routingContext){
+     return new ProxyContext(
+         RBAC_URI,
+         NO_HEADERS,
```
rohitsanj, line 196: Should we set `Content-Type: application/json` here instead of in `ControlPlaneAuthenticationProcessor`?

src/main/java/io/confluent/idesidecar/restapi/application/ProxyProcessorBeanProducers.java
```diff
@@ -37,13 +61,29 @@ public Processor<ClusterProxyContext, Future<ClusterProxyContext>> clusterProxyP
...
+  @Named("RBACProxyProcessor")
+  public Processor<ProxyContext, Future<ProxyContext>> RbacProxyProcessor(
+      RBACProxyProcessor rbacProxyProcessor
+  ) {
+    return Processor.chain(
+        new ConnectionProcessor<>(connectionStateManager),
+        controlPlaneAuthenticationProcessor,
+        rbacProxyProcessor,
+        new ProxyRequestProcessor(webClientFactory, vertx),
+        emptyProcessorProxyContext
```
rohitsanj, line 86: I think the ordering here needs an update: the `rbacProxyProcessor` sets the proxy request parameters, which should then be modified by the `controlPlaneAuthenticationProcessor` (instead of other way round).

src/main/java/io/confluent/idesidecar/restapi/resources/RestProxyResource.java
```diff
@@ -43,16 +57,65 @@ public void schemaRegistryClusterProxy(RoutingContext routingContext) {
...
+  @Route(
+      path = RBAC_RESOURCE_PATH,
+      methods = Route.HttpMethod.PUT,
+      produces = MediaType.APPLICATION_JSON,
+      consumes = MediaType.APPLICATION_JSON
+  )
+  @Blocking
+  public void RBACProxyRoute(RoutingContext routingContext) {
+    handleRBACProxy(routingContext, createRBACProxyContext(routingContext));
+  }
```
rohitsanj, line 76: It would be nice to add the request/response schema here so that it's part of the generated openapi spec. The way to do that would be to create Java `record` types that exactly represent the schema, and then specifying that here. 

cc @shouples

rohitsanj (changes requested):
@Cerchie, thanks! This looks great. Glad we were able to re-use the processor pattern for establishing the RBAC proxy endpoint. 

I have a few requested changes and minor comments.

src/main/java/io/confluent/idesidecar/restapi/resources/RestProxyResource.java
```diff
@@ -43,16 +64,78 @@ public void schemaRegistryClusterProxy(RoutingContext routingContext) {
...
+      @NotNull String title
+  ) {}
+
+  @Route(
+      path = RBAC_RESOURCE_PATH,
+      methods = Route.HttpMethod.PUT,
+      produces = MediaType.APPLICATION_JSON,
+      consumes = MediaType.APPLICATION_JSON
+  )
+
```
rohitsanj, line 101: Please remove this extra separation between the annotations.

src/main/java/io/confluent/idesidecar/restapi/resources/RestProxyResource.java
```diff
@@ -43,16 +64,78 @@ public void schemaRegistryClusterProxy(RoutingContext routingContext) {
...
+        @NotNull String resourceName,
+        @NotNull String operation,
+        @NotNull Scope scope
+    ) {
+      public record Scope(
+          @NotNull Map<String, String> clusters,
+          @NotNull String[] path
+      ) {}
+    }
+  }
```
rohitsanj, line 89: This looks great! I think we should remove the `@NotNull` constraints since we don't really utilize the request body anywhere but to generate the OpenAPI spec.

src/main/java/io/confluent/idesidecar/restapi/resources/RestProxyResource.java
```diff
@@ -43,16 +64,78 @@ public void schemaRegistryClusterProxy(RoutingContext routingContext) {
...
+
+  public record RBACResponse(
+      @NotNull String title
+  ) {}
```
rohitsanj, line 93: From the [ticket](https://github.com/confluentinc/ide-sidecar/issues/81), it looks like the response is simply a `String[]` and not an object with a field `title`.

src/main/java/io/confluent/idesidecar/restapi/resources/RestProxyResource.java
```diff
@@ -43,16 +64,78 @@ public void schemaRegistryClusterProxy(RoutingContext routingContext) {
...
+  @RequestBody(
+      description = "RBAC request body",
+      required = true,
+      content = @Content(schema = @Schema(implementation = RBACRequest.class))
+  )
+  @APIResponses({
+    @APIResponse(
+        responseCode = "200",
+        description = "Successful response",
+        content = @Content(schema = @Schema(implementation = RBACResponse.class))
```
rohitsanj, line 113: Should this be the following instead?

```suggestion
        content = @Content(schema = @Schema(implementation = String[].class))
```

src/main/java/io/confluent/idesidecar/restapi/resources/RestProxyResource.java
```diff
@@ -43,16 +64,78 @@ public void schemaRegistryClusterProxy(RoutingContext routingContext) {
...
+    @APIResponse(
+        responseCode = "401",
+        description = "Unauthorized",
+        content = @Content(schema = @Schema(implementation = RBACResponse.class))
+    ),
+    @APIResponse(
+        responseCode = "404",
+        description = "Not Found",
+        content = @Content(schema = @Schema(implementation = RBACResponse.class))
+    )
```
rohitsanj, line 124: I think we should remove these since we don't know exactly what the response schema would be for these error codes.

src/main/java/io/confluent/idesidecar/restapi/resources/RestProxyResource.java
```diff
@@ -43,16 +64,78 @@ public void schemaRegistryClusterProxy(RoutingContext routingContext) {
...
+  })
+  public void RBACProxyRoute(RoutingContext routingContext) {
```
rohitsanj, line 126: Let's use camelCase for method names.

```suggestion
  public void rbacProxyRoute(RoutingContext routingContext) {
```

src/main/java/io/confluent/idesidecar/restapi/resources/RestProxyResource.java
```diff
@@ -116,4 +181,14 @@ private ClusterProxyContext createSRClusterContext(RoutingContext routingContext
...
+    return new ProxyContext(
+        RBAC_URI,
+        NO_HEADERS,
+        HttpMethod.PUT,
+        routingContext.body().buffer(),
+        NO_PATH_PARAMS,
+        routingContext.request().getHeader(RequestHeadersConstants.CONNECTION_ID_HEADER)
+    );
+  }
+}
```
rohitsanj, line 194: Please add newline to end of file.

src/generated/resources/openapi.json
```diff
@@ -33,6 +33,15 @@
...
     },
+    "/api/metadata/security/v2alpha1/authorize" : {
+      "put" : {
+        "responses" : {
+          "204" : {
+            "description" : "No Content"
+          }
+        }
+      }
+    },
```
rohitsanj, line 44: Can you please try running `make clean build` and see if this section is updated to include the request and response schemas? (Same for openapi.yaml)

rohitsanj (changes requested):
This is looking great and is very close to being done. I've requested a few more changes, please have a look. Thanks!

src/main/java/io/confluent/idesidecar/restapi/resources/RestProxyResource.java
Cerchie: Hm yeah, removed it since we can just use a String[] later

rohitsanj (approved):
LGTM! Thanks for working through the changes on this. Tested locally and it works great.

PR #328, opened 2025-02-14
Use SchemaRegistryConfig for local connections

rohitsanj:
## Summary of Changes

- Instructs RealLocalFetcher to look up SR URI in the `schema_registry.uri` config, and if not present, fallback to `local_config.schema-registry-uri`
- Behavior change: We no longer default to the SR URI of `localhost:8081` if `local_config` (or `local_config.schema-registry-uri`) is null.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
@@ -153,12 +153,17 @@ public class RealLocalFetcher extends ConfluentLocalRestClient implements LocalF
   }
 
   public String resolveSchemaRegistryUri(String connectionId) {
-    var localConfig = connections.getConnectionSpec(connectionId).localConfig();
-    if (localConfig == null || localConfig.schemaRegistryUri() == null) {
-      // Use the default SR URI when missing LocalConfig or missing SR URI in that config
-      return CONFLUENT_LOCAL_DEFAULT_SCHEMA_REGISTRY_URI;
+    var localSpec = connections.getConnectionSpec(connectionId);
+
+    String uri;
+    if (localSpec.schemaRegistryConfig() != null) {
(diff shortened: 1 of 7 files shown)


rohitsanj:
From Copilot:

> [nitpick] The method distinguishes between a null URI (returning the default) and a blank URI (returning null). Please confirm that this dual behavior is intentional to avoid unexpected outcomes.

For human PR reviewers, this is totally intentional indeed. 😆

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
```diff
@@ -76,13 +76,6 @@ public class RealLocalFetcher extends ConfluentLocalRestClient implements LocalF
           String.class
       );
 
-  static final String CONFLUENT_LOCAL_DEFAULT_SCHEMA_REGISTRY_URI = ConfigProvider
-      .getConfig()
-      .getValue(
-          "ide-sidecar.connections.confluent-local.default.schema-registry-uri",
```
shouples, line 82: If this is removed, does anything else have to use https://github.com/confluentinc/ide-sidecar/blob/main/src/main/resources/application.yml#L126?

shouples (approved):
LGTM and native executable works with the extension! ✅ 
Just had one question about the config defaults, but shouldn't block this

src/main/java/io/confluent/idesidecar/restapi/models/graph/RealLocalFetcher.java
rohitsanj: Good point, there's nothing having to use that. Will clean it up once we remove `LOCAL` connections for good.

PR #332, opened 2025-02-21
MINOR: use sidecar version as OpenAPI spec version

shouples:
Removes the hard-coded OpenAPI spec `1.0.1` in favor of the quarkus app version, so anytime the sidecar version changes, it should also be reflected in the OpenAPI spec.

(This should also help on the VS Code extension side to avoid bumping the version without the associated spec version.)


rohitsanj (approved):
LGTM!

PR #335, opened 2025-02-28
Add CCloud Produce API

rohitsanj:
## Summary of Changes

Since the Confluent Cloud Produce Records API only supports passing data in type `JSON` or `BINARY`. The `JSON` type represents a schemaless JSON blob (string, object, list, number, etc) and the `BINARY` type represents a base64-encoded string of the serialized bytes. 

For vscode to produce schematized records to Confluent Cloud Kafka topics, it would've had to re-implement the serialization logic, each for avro, protobuf and jsonschema. Given there's [precedence](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/kafkarest/RecordSerializer.java) for this serialization logic in the sidecar's "internal" Produce Records API (used for direct Kafka connections), it would be prudent to re-use that logic and employ it in a separate route meant for Confluent Cloud Produce. This new route (`/gateway/v1/clusters/{cluster_id}/topics/{topic_name}/records`) is exposed under `/gateway/v1/` and is NOT part of the Kafka REST spec, but is closely modeled after it. 

The route does the following:

1. **Client sends a request** to the sidecar with the key/value data and (optionally) schema parameters for each.
1. **Serializes the key and value data** based on whether schema parameters were set (`subject`, `subject_name_strategy` and `schema_version`) -- if schema is set, we serialize the data into bytes using the appropriate serializer, and then base64 encode. If schema is not set, meaning it's just JSON, we send it over as is. (Note: the implementation uses a `KafkaJsonDeserializer` to deserialize the already-serialized JSON to get us.. back to JSON, but this detail can be ignored when looking at high level of the endpoint).
1. **Sends off a request to the Confluent Cloud Produce Records API** using the data plane token for authentication
1. **Parse the response and return it to the client**

Many existing code paths have been re-used or re-purposed to implement the CCloud Produce endpoint. A lot of the changes are just refactors. Net new code added is ~450 lines of code (excluding openapi changes). 

Will add a few tests soon.

All existing tests pass, doesn't prove absence of bugs but no regressions guaranteed at the minimum. 

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


Cerchie:
question: do we need the subject name strategy `ProduceRequestData` if VSCode will not use it but use subject instead?

rohitsanj:
> question: do we need the subject name strategy ProduceRequestData if VSCode will not use it but use subject instead?

I think it doesn't hurt, so yes. The current implementation in the sidecar expects it to be passed and I wouldn't want to change that for now.

shouples (approved):
Click-tested with https://github.com/confluentinc/vscode/pull/1093 and works great for CCloud and non-CCloud producing with schemas! 🚀

PR #345, opened 2025-03-06
Include schema details in consume response and support producing/consuming raw bytes

### LINKED ISSUE #117: Return base64 encoded bytes under `__raw__` in Consume API responses
With the way the sidecar consume API is implemented today, the VS Code message viewer and the user do not have a way to determine whether the key/value data returned by the sidecar Consume API is either (a) a string representation of a string literal `"i am a teapot"`, or (b) a string representation of the base64 encoded version of some raw bytes `"aSBhbSBub3QgYSB0ZWFwb3QK"` (decodes to "i am not a teapot").

The CCloud API handles this by sending the base64 encoded data in the `__raw__` field of the response.  We need to consider implementing this for the sidecar Consume API as well.

### LINKED ISSUE #343: Consume messages route result needs to describe per-record deser metadata
Need to have consume message route include additional metadata fields describing how each consumed record was deserialized:
  * The key schema id used, or explicit null if none.
  * The value schema id used, or explicit null if none.
  * The deserializer tech used: enum of AVRO, PROTOBUF, JSONSCHEMA, or null if no deser done and presenting raw bytes.

Intent is to be able to describe to vscode message viewer the details about the deser used so could differentate between schemaless raw messages vs those actually deserialized and through what means. Message viewer could then link directly to either schema, ... .

rohitsanj:
## Summary of Changes

- Added `key_schema` and `value_schema` fields to the Consume API response, has two fields `schema_id` and `deserializer_tech`. Refer openapi spec for how this looks.
- In order to test the `RAW_BYTES` enum value being returned in the consume response, I went ahead and implemented support for producing/consuming raw bytes which is something we intended to do (see #117).
- Fixed many tests -- it's a good signal that a change like this breaks many (but not too many) tests meaning our coverage is sufficient.

Resolves #343 
Resolves #117

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


src/main/java/io/confluent/idesidecar/restapi/messageviewer/data/SimpleConsumeMultiPartitionResponse.java
```diff
@@ -59,39 +60,21 @@ public record ExceededFields(
...
       @JsonProperty("key") JsonNode key,
       @JsonProperty("value") JsonNode value,
+      @JsonProperty("key_schema") SchemaDetails keySchema,
+      @JsonProperty("value_schema") SchemaDetails valueSchema,
       @JsonProperty("key_decoding_error") String keyDecodingError,
       @JsonProperty("value_decoding_error") String valueDecodingError,
       @JsonProperty("exceeded_fields") ExceededFields exceededFields
   ) {
-    // Initialize key and value decoding errors to null by default
-    public PartitionConsumeRecord(
```
rohitsanj, line 74: unused

src/generated/resources/openapi.yaml
```diff
@@ -1023,6 +1047,10 @@ components:
           $ref: "#/components/schemas/JsonNode"
         value:
           $ref: "#/components/schemas/JsonNode"
+        key_schema:
+          $ref: "#/components/schemas/SchemaDetails"
+        value_schema:
+          $ref: "#/components/schemas/SchemaDetails"
```
flippingbits, line 1053: Since these keys might also hold information about keys/values without a schema, what about renaming them to `key_format` and `value_format`?

src/generated/resources/openapi.yaml
```diff
@@ -741,6 +741,30 @@ components:
       format: date
       type: string
       example: 2022-03-10
+    DeserializerTech:
```
flippingbits, line 744: I feel this enum models [(data) formats](https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/index.html#supported-formats) instead of something specific to deserialization. WDYT?

flippingbits (commented):
Thanks, @rohitsanj. This looks great! I have only a few minor questions.

src/generated/resources/openapi.json
```diff
@@ -1900,16 +1931,16 @@
...
           "data" : {
             "type" : "object",
             "nullable" : true
           },
           "status" : {
             "enum" : [ "UP", "DOWN" ],
             "type" : "string"
+          },
+          "name" : {
+            "type" : "string"
```
rohitsanj, line 1943: I've given up with these things. Please ignore it.

rohitsanj:
Tested that message viewer still functions well.

Screenshot 2025-03-07 at 10 53 21 AM

src/main/java/io/confluent/idesidecar/restapi/messageviewer/RecordDeserializer.java
```diff
@@ -389,48 +420,77 @@ private static Optional<DecodedResult> maybeTrivialCase(
...
+  private static WrappedJson safeRead(byte[] bytes) {
     try {
-      return OBJECT_MAPPER.readTree(bytes);
+      return new WrappedJson(
+          OBJECT_MAPPER.readTree(bytes),
+          DataFormat.JSON
+      );
     } catch (IOException e) {
-      return TextNode.valueOf(new String(bytes, StandardCharsets.UTF_8));
+      return new WrappedJson(ByteArrayJsonUtil.asJsonNode(bytes), DataFormat.RAW_BYTES);
```
rohitsanj, line 487: Note: This is a major behavior change. We interpret bytes as base64 encoded string instead of UTF-8.

src/test/java/io/confluent/idesidecar/restapi/kafkarest/RecordsV3BaseSuite.java
```diff
@@ -279,7 +333,7 @@ private static Set<ProduceRequestHeader> convertResponseHeaders(
         .collect(Collectors.toSet());
   }
 
-  default void assertSame(JsonNode actual, Object expected) {
+  default void assertSame(Object expected, JsonNode actual) {
```
rohitsanj, line 336: This changed line is the JUnit convention, so was annoying to have it the other way around.

flippingbits (approved):
LGTM, pending a passing CI pipeline. Thanks, @rohitsanj!

PR #387, opened 2025-03-31
README [ci skip]: Add samples for control plane proxy 

Cerchie:
## Summary of Changes

Updates CONTRIBUTING.md with example of how to hit the artifacts and fcpm compute pool endpoints. 

cc: @shouples

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -690,9 +690,7 @@ query localConnections {
 curl -s -H "Content-Type:application/json" -H "Authorization: Bearer ${DTX_ACCESS_TOKEN}" http://localhost:26636/gateway/v1/graphql/schema.graphql
 ```
 
-### Invoking the Control Plane Proxy API 
-
-#### Examples For Artifact and Compute Pool Management with Flink
+### Invoking the Control Plane Proxy API
 
 The Control Plane Proxy API allows you to make calls to the Confluent Cloud control plane, specifically for artifact and compute pool management endpoints defined by the configuration:
 
@@ -715,6 +713,8 @@ Required headers:
 
 > Notes:
   Control plane endpoints require a valid control plane token which is automatically managed by the proxy
+
+#### Examples For Artifact and Compute Pool Management with Flink
 
 ##### Example: List Compute Pools
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -690,7 +690,9 @@ query localConnections {
 curl -s -H "Content-Type:application/json" -H "Authorization: Bearer ${DTX_ACCESS_TOKEN}" http://localhost:26636/gateway/v1/graphql/schema.graphql
 ```
 
-### Invoking the Control Plane Proxy API For Artifact and Compute Pool Management
+### Invoking the Control Plane Proxy API 
+
+#### Examples For Artifact and Compute Pool Management with Flink
 
 The Control Plane Proxy API allows you to make calls to the Confluent Cloud control plane, specifically for artifact and compute pool management endpoints defined by the configuration:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -690,6 +690,45 @@ query localConnections {
 curl -s -H "Content-Type:application/json" -H "Authorization: Bearer ${DTX_ACCESS_TOKEN}" http://localhost:26636/gateway/v1/graphql/schema.graphql
 ```
 
+### Invoking the Control Plane Proxy API
+
+The Control Plane Proxy API allows you to make calls to the Confluent Cloud control plane, specifically for artifact and compute pool management endpoints defined by the configuration:
```
shouples, line 695: > specifically for artifact and compute pool management endpoints
Will this section have to be updated every time something is added to the Control Plane Proxy regex pattern? Or is this section just an example focusing on the Flink routes?

CONTRIBUTING.md
```diff
@@ -690,6 +690,45 @@ query localConnections {
...
+
+```yaml
+ccloud-api-control-plane-regex: "(/artifact.*)|(/fcpm/v2/compute-pools.*)"
+```
```
flippingbits, line 699: Should we add a link to the application.yml so that folks can figure out the full config path, `ide-sidecar.connections.ccloud.control-plane-base-url`?

CONTRIBUTING.md
```diff
@@ -690,6 +690,45 @@ query localConnections {
...
+
+Control Plane endpoints:
+  http://localhost:26636/artifact/* (For artifact management)
+  http://localhost:26636/fcpm/v2/compute-pools/* (For Flink compute pools management)
+
+Required headers:
+
+- Authorization: Bearer ${DTX_ACCESS_TOKEN}
+- x-connection-id: <connection-id> (Must reference a valid Confluent Cloud connection)
```
flippingbits, line 708: Should we say "authenticated" Confluent Cloud connection? Otherwise, the connection won't hold a (valid) control plane token.

```suggestion
- x-connection-id: <connection-id> (Must reference an authenticated Confluent Cloud connection)
```

CONTRIBUTING.md
```diff
@@ -690,6 +690,45 @@ query localConnections {
...
+> Notes:
+  Control plane endpoints require a valid control plane token which is automatically managed by the proxy
+
+#### Example: List Compute Pools
+
+```bash
+❯ curl -s \
+  -H "Authorization: Bearer ${DTX_ACCESS_TOKEN}" 
+  -H "x-connection-id: c1" 
+  http://localhost:26636/fcpm/v2/compute-pools\?environment\=ccloud-dev | jq -r .
```
flippingbits, line 719: What about adding an example response to this snippet?

CONTRIBUTING.md
flippingbits: I think it'd be great to highlight that we can easily add more control plane routes to the proxy by updating the config below.

flippingbits (commented):
Thanks, @Cerchie! I have a few suggestions.

CONTRIBUTING.md
Cerchie: Good question... I would _like_ for that to happen haha but I don't know that we have a way of enforcing that right now.

shouples: Maybe keep the top-level header and description generic, and use a sub-header for something like "Example: Managing Flink Compute Pools & Artifacts"?

Cerchie: oooh good idea, pushing that up

flippingbits (approved):

PR #390, opened 2025-04-01
Migrate RBAC proxy to generic control plane proxy

eguicf:
## Summary of Changes

Migrated the existing [RBAC proxy](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/application/ProxyProcessorBeanProducers.java) (which is very specific to the proxied CCloud API endpoint) to the new generic control plane proxy.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


Cerchie (changes requested):
Thanks Ellie, & congrats on your 1st PR to the sidecar! 🥇 

We should be able to delete the `RBACProxyProcessor` class now, since it seems these were the only usages.

rohitsanj:
I think we also update the config in application.yml to include to RBAC endpoint path?

src/main/resources/application.yml
```diff
@@ -159,7 +159,7 @@ ide-sidecar:
     # How often to check?
     interval-seconds: 60
   proxy:
-    ccloud-api-control-plane-regex: "(/artifact.*)|(/fcpm/v2/compute-pools.*)"
+    ccloud-api-control-plane-regex: "(/artifact.*)|(/fcpm/v2/compute-pools.*)|(/api/metadata/security/v2alpha1/authorize)"
```
rohitsanj, line 162: If the base url is `https://api.confluent.cloud` -- then the `/api/metadata/..` probably won't work. Can we check to see if `https://api.confluent.cloud/metadata/security/v2alpha1/authorize` is a valid way to call the endpoint? If so, let's do:
```suggestion
    ccloud-api-control-plane-regex: "(/artifact.*)|(/fcpm/v2/compute-pools.*)|(/metadata/security/v2alpha1/authorize)"
```

rohitsanj (changes requested):
Thanks, @eguicf! This looks good, I just have one comment. Let me know if you'd like to pair on it.

src/main/resources/application.yml
eguicf: thank you for the comment! @rohitsanj. I have update that. but currently still fixing some test errors. will let you know!

Issue #393, opened 2025-04-03
Validate ccloud region and provider for Flink routes


Cerchie:
from @rohitsanj :another thing we can address in a follow up is validating that ccloud region and provider combination is valid. For instance, (us-east-200, linode) should fail.

Cerchie:
closed via #397

PR #397, opened 2025-04-04
#393: Verify provider

Cerchie:
Getting a start on #393 and validating the provider. No enum exists for regions like it does with providers, so I focused on the provider for now and just added an empty check for the region. There does happen to be [a list](https://docs.confluent.io/cloud/current/clusters/regions.html) if we decide to implement something. 

## Pull request checklist

- Tests:
    - [x] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```


flippingbits (commented):
Great work, @Cerchie! I have one comment on the returned error message.

PR #402, opened 2025-04-09
Proxy for Flink Language Service

flippingbits:
## Summary of Changes

This PR adds the WebSockets-based proxy for the Flink Language Service. It listens on the path `/flsp` and allows the extension to interact with the CCloud Flink Language Service without having to manage authentication tokens or handle reconnections.

The proxy is designed as a _simple_ pass-through proxy. When the extension connects to it, the proxy will establish a WebSockets connection to the CCloud Flink Language Service and send an initial authentication message. From this point on, it will pass all messages sent by the extension as they are to the Flink Language Service and vice versa.

When the connection between the proxy and the Flink Language Service gets closed, e.g., because the data plane token has expired, the proxy will attempt up to `MAX_RECONNECT_ATTEMPTS` (default: 1) reconnection attempts before closing the connection with the extension. Every message that was sent by the Flink Language Service and received by the proxy resets the reconnect attempts counter.

Sometimes, the Language Services requires the extension to pass an access token. To this end, the proxy replaces all occurrences of the placeholder `{{ ccloud.data_plane_token }}` with the current data plane token. Note that this is a very simple regex-powered string replacement, not a proper template engine -- please make sure to use the exact notation for the placeholder.

When connecting, the extension must provide a few information via query params to set the proxy to the CCloud Language Service up:

* the query param `connectionId` must point to an authenticated connection of type `CCLOUD`
* the query param `region` must point to the Flink region of the Language Service, e.g., `us-east-1`
* the query param `provider` must point to the Flink provider of the Language Service, e.g., `gcp`
* the query param `environmentId` must point to a CCloud environment
* the query param `organizationId` must point to a CCloud organization

Note that the sidecar also requires the extension to pass the access token via the request header `Authorization`, as for most other endpoints.

The following example shows how to connect to the proxy using the CLI tool [wscat](https://github.com/websockets/wscat):

```shell
wscat -H"Authorization: Bearer $access_token" -c"ws://127.0.0.1:26636/flsp?connectionId=1&region=us-east1&provider=gcp&environmentId=env-id&organizationId=org-id"
```

The PR is still in a very early state. I want to work on a couple of items before marking it as ready to be reviewed:

- [x] Make settings, like the [maximum number of reconnect attempts](https://github.com/confluentinc/ide-sidecar/blob/language-service-ws-proxy/src/main/java/io/confluent/idesidecar/websocket/resources/FlinkLanguageServiceProxy.java#L31) and the [language service's URL](https://github.com/confluentinc/ide-sidecar/blob/language-service-ws-proxy/src/main/java/io/confluent/idesidecar/websocket/resources/FlinkLanguageServiceProxy.java#L32), configurable via the [application.yml](https://github.com/confluentinc/ide-sidecar/blob/language-service-ws-proxy/src/main/resources/application.yml)
- [x] Split the class [FlinkLanguageServiceProxy](https://github.com/confluentinc/ide-sidecar/blob/language-service-ws-proxy/src/main/java/io/confluent/idesidecar/websocket/resources/FlinkLanguageServiceProxy.java) into multiple files
- [x] Support injecting control plane token
- [x] General hardening and more sophisticated error handling
- [x] Mock the CCloud Flink Language Service and implement integration tests
- [x] Documentation

## Pull request checklist

- Tests:
    - [x] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/websocket/proxy/ProxyContext.java
+++ b/src/main/java/io/confluent/idesidecar/websocket/proxy/ProxyContext.java
@@ -1,0 +1,41 @@
+package io.confluent.idesidecar.websocket.proxy;
+
+import jakarta.websocket.Session;
+import org.eclipse.microprofile.config.ConfigProvider;
+
+public record ProxyContext (
+    String connectionId,
+    String region,
+    String provider,
+    String environmentId,
+    String organizationId
+) {
+
+  static final String LANGUAGE_SERVICE_URL_PATTERN = ConfigProvider
+      .getConfig()
+      .getValue("ide-sidecar.flink-language-service-proxy.url-pattern", String.class);
+
+  static final String CONNECTION_ID_PARAM_NAME = "connectionId";
+  static final String REGION_PARAM_NAME = "region";
+  static final String PROVIDER_PARAM_NAME = "provider";
+  static final String ENVIRONMENT_ID_PARAM_NAME = "environmentId";
+  static final String ORGANIZATION_ID_PARAM_NAME = "organizationId";
+
+  public static ProxyContext from(Session session) {
+    var paramMap = session.getRequestParameterMap();
+    return new ProxyContext(
+        paramMap.get(CONNECTION_ID_PARAM_NAME).get(0),
+        paramMap.get(REGION_PARAM_NAME).get(0),
+        paramMap.get(PROVIDER_PARAM_NAME).get(0),
+        paramMap.get(ENVIRONMENT_ID_PARAM_NAME).get(0),
+        paramMap.get(ORGANIZATION_ID_PARAM_NAME).get(0)
+    );
+  }
+
+  public String getConnectUrl() {
+    // TODO: I guess this won't work for private networks, we'll need something more sophisticated
+    return LANGUAGE_SERVICE_URL_PATTERN
+        .replace("{{ region }}", region)
+        .replace("{{ provider }}", provider);
+  }
+}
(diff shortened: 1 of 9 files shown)


src/main/java/io/confluent/idesidecar/websocket/proxy/ProxyContext.java
```diff
@@ -0,0 +1,55 @@
...
+        region,
+        provider,
+        environmentId,
+        organizationId,
+        connection
+    );
+  }
+
+  public String getConnectUrl() {
+    // TODO: I guess this won't work for private networks, we'll need something more sophisticated
```
noeldevelops, line 50: 👍 
It looks like in the UI they are turning off autocomplete for private endpoints for now:
https://github.com/confluentinc/frontend-vault/blob/master/packages/workbench/src/components/Editor/useAutocompleteClientRef.ts#L35-L46

noeldevelops (commented):
This looks like a great start to me @flippingbits. I don't imagine we'll need to add much more, but might have to make some changes to close/reconnect logic depending on how we decide to handle it vscode-side. 

I am able to connect (including the auth headers) in [my local branch](https://github.com/confluentinc/vscode/pull/1502) and get the completions and diagnostics. 🙌

rohitsanj (approved):
Fantastic work, @flippingbits! Clean and easy to read. LGTM

PR #443, opened 2025-06-02
feat: Always check connection status upon creation

### LINKED ISSUE #442: Make sure we always check the health of connections upon creation
When hitting the endpoint `POST /gateway/v1/connections` and not providing the `dryRun=true` query param, checking the status of the new connection might be performed with quite some delay because it's performed by the [recurring health check](https://github.com/confluentinc/ide-sidecar/blob/main/src/main/java/io/confluent/idesidecar/restapi/connections/RefreshConnectionStatuses.java). This will become problematic if we increase the interval for the recurring health check (as done by https://github.com/confluentinc/ide-sidecar/pull/436 for direct connections).

Thus, we should make sure we always check the health of connections upon creation. The health check should be performed asynchronously unless the query param `dryRun` equals `true`.

flippingbits:
## Summary of Changes

This change makes sure we always check the connection status immediately after creating it, so that we don't run into issues when increasing the interval for the recurring health status check.

Fixes #442

## Pull request checklist

- Tests:
    - [ ] Added new
    - [ ] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/resources/ConnectionsResource.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/resources/ConnectionsResource.java
@@ -14,6 +14,7 @@ import io.confluent.idesidecar.restapi.models.ConnectionsList;
 import io.quarkus.logging.Log;
 import io.smallrye.common.annotation.Blocking;
 import io.smallrye.mutiny.Uni;
+import io.vertx.core.Vertx;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.DELETE;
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/idesidecar/restapi/resources/ConnectionsResource.java
```diff
@@ -90,12 +94,15 @@ public Uni<Connection> createConnection(
               Connection.from(testedState, connectionStatus)
           );
     }
-    // Create a real connection state and return the connection using the latest status
-    var newState = connectionStateManager.createConnectionState(connectionSpec);
+    // Create the connection
+    var connection = connectionStateManager.createConnectionState(connectionSpec);
+    // Check the connection's status without blocking on it
```
jlrobins, line 99:
```suggestion
    // Immediately kick off async check of the new connection, independent of the periodic scheduled task, which may not fire for "a while" from now. Overlapping checks at connection creation time are fine and should resolve to the same state. 
```

jlrobins (approved):

PR #449, opened 2025-06-04
fix: Specify that `directConnectionById` may return `null`

flippingbits:
## Summary of Changes

This change updates the GraphQL schema to specify that the query `directConnectionById` may return a `null` value. Also, it performs a few minor refactorings.

## Pull request checklist

- Tests:
    - [ ] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/main/java/io/confluent/idesidecar/restapi/resources/DirectQueryResource.java
+++ b/src/main/java/io/confluent/idesidecar/restapi/resources/DirectQueryResource.java
@@ -44,9 +44,8 @@ public class DirectQueryResource {
    */
   @Query("directConnectionById")
   @Description("Get direct connection by ID")
-  @NonNull
-  public DirectConnection getDirectConnectionByID(String connectionID) throws Exception {
-    return direct.getDirectConnectionByID(connectionID);
+  public Uni<DirectConnection> getDirectConnectionByID(@NonNull String id) {
+    return direct.getDirectConnectionByID(id);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/idesidecar/restapi/resources/DirectQueryResource.java
```diff
@@ -44,9 +44,8 @@ public Uni<List<DirectConnection>> getDirectConnections() {
    */
   @Query("directConnectionById")
   @Description("Get direct connection by ID")
-  @NonNull
```
flippingbits, line 47: The removal of the above line is the meat of this PR.

src/main/java/io/confluent/idesidecar/restapi/resources/DirectQueryResource.java
```diff
@@ -44,9 +44,8 @@ public Uni<List<DirectConnection>> getDirectConnections() {
...
-  public DirectConnection getDirectConnectionByID(String connectionID) throws Exception {
-    return direct.getDirectConnectionByID(connectionID);
+  public Uni<DirectConnection> getDirectConnectionByID(@NonNull String id) {
```
flippingbits, line 47: Adding `@NonNull` to the parameter `id` because we don't want to accept `null` values.

PR #461, opened 2025-07-09
Allow GraphQL API to return partial data/failure responses

### LINKED ISSUE #460: Allow GraphQL API to return partial data/failure responses
At the moment, the GraphQL API fails completely if fetching Kafka clusters or Flink compute pools fails.
We should allow the GraphQL API to return partial data in case of a partial failure, so that the VS Code extension can, for instance, list Kafka clusters when the API endpoint for listing Flink compute pools is broken.

flippingbits:
## Summary of Changes

This change allows the GraphQL API to return partial data in case of a partial failure, so that the VS Code extension can, for instance, list Kafka clusters when the API endpoint for listing Flink compute pools is broken.

Fixes #460

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [ ] Deleted existing
    ```shell
    make quarkus-dev
    ```
    ```shell
    ```

### DIFF
--- a/src/generated/resources/schema.graphql
+++ b/src/generated/resources/schema.graphql
@@ -8,10 +8,10 @@ type CCloudConnection {
 
 type CCloudEnvironment {
   connectionId: String!
-  flinkComputePools: [CCloudFlinkComputePool]!
+  flinkComputePools: [CCloudFlinkComputePool]
   governancePackage: CCloudGovernancePackage!
   id: String!
-  kafkaClusters: [CCloudKafkaCluster]!
+  kafkaClusters: [CCloudKafkaCluster]
   name: String!
   organization: CCloudReference!
   schemaRegistry: CCloudSchemaRegistry
@@ -115,9 +115,9 @@ type Query {
   "Get all direct connections"
   directConnections: [DirectConnection]!
   "Find CCloud Kafka clusters using a connection and various criteria"
-  findCCloudKafkaClusters(connectionId: String!, environmentId: String = "", name: String = "", provider: String = "", region: String = ""): [CCloudKafkaCluster]!
+  findCCloudKafkaClusters(connectionId: String!, environmentId: String = "", name: String = "", provider: String = "", region: String = ""): [CCloudKafkaCluster]
   "Get Flink compute pools for a specific connection and environment"
-  getFlinkComputePools(connectionId: String!, envId: String!): [CCloudFlinkComputePool]!
+  getFlinkComputePools(connectionId: String!, envId: String!): [CCloudFlinkComputePool]
   "Get all local connections"
   localConnections: [LocalConnection]!
 }
--- a/src/test/resources/graph/real/get-ccloud-connection-by-id-failed-compute-pool-expected.json
+++ b/src/test/resources/graph/real/get-ccloud-connection-by-id-failed-compute-pool-expected.json
@@ -1,0 +1,83 @@
+{
+  "errors": [
+    {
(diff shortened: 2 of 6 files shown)


src/test/resources/graph/real/get-ccloud-connection-by-id-failed-kafka-expected.json
```diff
@@ -50,7 +50,16 @@
       ],
       "environments":
       [
-        null,
+        {
+          "id": "env-x7727g",
+          "name": "main-test-env",
+          "kafkaClusters": null,
```
flippingbits, line 56: Set to `null` because fetching Kafka clusters failed for this Environment.

src/test/resources/graph/real/get-ccloud-connection-by-id-failed-compute-pool-expected.json
```diff
@@ -0,0 +1,83 @@
...
+            "uri": "https://psrc-ghijk0.us-west2.gcp.confluent.cloud"
+          },
+          "flinkComputePools": null
+        },
+        {
+          "id": "env-kkk3jg",
+          "name": "env-with-no-kafka-cluster",
+          "kafkaClusters": [],
+          "schemaRegistry": null,
+          "flinkComputePools": []
```
flippingbits, line 78: Set to an empty array (`[]`) because this Environment does not hold any Flink Compute Pools. It allows the extension to differentiate between failures and empty Environments. Additionally, the extension should investigate the `errors` field.

src/generated/resources/schema.graphql
```diff
@@ -115,9 +115,9 @@ type Query {
   "Get all direct connections"
   directConnections: [DirectConnection]!
```
Cerchie, line 116: I see that other resources, such as Direct Connections, are still required? is it fairly certain we won't run into a similar issue with them?

Cerchie (approved):
One comment because I was curious about something, but I verified the change locally and it returned a good response: 

```bash
❯ curl -s -X POST "http://localhost:26636/gateway/v1/graphql" \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer ${DTX_ACCESS_TOKEN}" \
  -d '{
    "query": "query devConnection {\n  ccloudConnectionById(id: \"ccloud-dev\") {\n    id\n    name\n    type\n    organizations {\n      id\n      name\n      current\n    }\n    environments {\n      id\n      name\n      kafkaClusters {\n        id\n        name\n        provider\n        region\n        bootstrapServers\n        uri\n      }\n      schemaRegistry {\n        provider\n        region\n        uri\n      }\n      flinkComputePools {\n        id\n        display_name\n        provider\n        region\n        max_cfu\n      }\n    }\n  }\n}"
  }' | jq -r .
{
  "data": {
    "ccloudConnectionById": {
      "id": "ccloud-dev",
      "name": "Dev Connection",
      "type": "CCLOUD",
      "organizations": [
        {
          "id": "ORGID",
          "name": "NAME,
          "current": true
        }
      ],
      "environments": [
        {
```

src/generated/resources/schema.graphql
flippingbits: Good question! Yes, I think we should keep the _top-level_ resources, like Connections or CCloud Environments, required. First, we manage connections internally and don't need to consult external APIs for fetching them. Secondly, if we are not able to read connections, it does not make sense to fetch subresources, like Kafka clusters.

PR #484, opened 2025-08-07
fix(CI): add `--force` to overwrite existing native executable artifacts in CI

shouples:
We hit a [503 error while trying to build a native executable](https://semaphore.ci.confluent.io/jobs/f6f465f0-5ba3-4303-bdeb-4dc243bf48c2) before uploading to the GitHub release, but unfortunately we couldn't re-run `multi-arch-builds-and-upload` since the artifacts were already pushed for that [workflow](https://semaphore.ci.confluent.io/workflows/5e640e9f-b300-4a9d-b765-552c9c901395?pipeline_id=6793f328-fc24-469e-a286-8824615e9e06). (Note: MacOS AMD64 successfully built and pushed the native executable artifact in that workflow, but only because it was the one that 503'd in the previous run.)

[Non-Windows](https://semaphore.ci.confluent.io/jobs/c2ee84a8-2e93-49ed-b112-6e11d3f784b6) (strangely didn't fail the job):
image

[Windows](https://semaphore.ci.confluent.io/jobs/b312ae7a-775f-4735-ab00-10b7a40d914f):
image

This change just adds the `--force` flag so any lingering artifacts are overwritten during the push.

### DIFF
--- a/.semaphore/build_native_executable_epilogue_macos_linux.sh
+++ b/.semaphore/build_native_executable_epilogue_macos_linux.sh
@@ -2,4 +2,4 @@ make store-test-results-to-semaphore
 make ci-bin-sem-cache-store
 make cache-docker-images
 NATIVE_EXECUTABLE=$(find target -name "*-runner")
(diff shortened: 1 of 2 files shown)


flippingbits (approved):
LGTM, thanks!

PR #485, opened 2025-08-08
use fossa installed in semaphoare agent

lyao-77:
since we installed fossa binary in the semaphore agent already, the installation of fossa can be removed now.


flippingbits (approved):
LGTM, thanks!

We used an older version (`3.6.8`) than the one installed on Semaphore (`3.10.14`). Let's see if it causes any issues.

PR #551, opened 2026-07-31
feat: identify failed record serialization parts

### LINKED ISSUE #416: Need to indicate if message produce schema serialization issue happend on key vs value schema
When produce / serialization error happens, the exception returned to extension side needs to indicate if the issue encountered with the key schema or the value schema. Please include some machine-readable clear attribute like "messagePart: [key|value]" or whatnot in the exception JSON.

Knowing which will help extension indicate to the user, say, through only marking either the key or value subset of the produced message as having error.

Pybsama:
## Summary of Changes

- add an optional `message_part` field to produce serialization errors, with `key`, `value`, and `both` values
- preserve the existing error message and ordinary bad-request response shape
- document the generated response model and register it for native reflection

The key and value serializers now retain which side failed through the parallel dry-run pipeline. If both serializers fail, the response reports `both`; unrelated 400 responses continue to omit `message_part`.

Local validation completed:

- `./mvnw --settings .mvn/settings.xml -Dtest=io.confluent.idesidecar.restapi.kafkarest.GenericProduceRecordTest -DskipITs verify`
- `make mvn-generate-sidecar-openapi-spec`
- `make mvn-package-native-sources-only`
- generated native reflection configuration inspection

The full non-IT unit run passed 710 of 713 tests. The remaining three `PreferencesResourceTest` cases also fail in isolation because their test resource URLs are percent-encoded under this non-ASCII checkout path and are then treated as filesystem paths. Container-backed integration tests were not run locally because Docker is unavailable; CI can exercise the added `RecordsV3ErrorsSuite` assertions.

Closes #416

## Pull request checklist

- Tests:
    - [x] Added new
    - [x] Updated existing
    - [ ] Deleted existing

### DIFF
--- a/kafka-rest.openapi.yaml
+++ b/kafka-rest.openapi.yaml
@@ -2256,6 +2256,19 @@ components:
           type: string
           nullable: true
 
+    ProduceRecordError:
+      allOf:
+        - $ref: '#/components/schemas/Error'
+        - type: object
+          properties:
+            message_part:
+              type: string
+              description: The part of the record that failed local serialization.
+              enum:
+                - key
+                - value
+                - both
+
     PartitionData:
       allOf:
         - $ref: '#/components/schemas/Resource'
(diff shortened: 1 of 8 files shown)


Pybsama:
Revalidated the exact PR head `7b586df` against the still-current upstream base `89f6367` on macOS/arm64 with OpenJDK 21.0.12:

- `GenericProduceRecordTest`: 4/4 pass via `verify -DskipITs=true`; Checkstyle and SpotBugs pass (0 findings).
- `make mvn-generate-sidecar-openapi-spec`: pass; generated specs leave the tracked tree clean.
- `make mvn-package-native-sources-only`: pass; the native runner's `reflect-config.json` contains `ProduceRecordError`, `MessagePartEnum`, and both builder classes.
- Full non-IT unit suite: 710/713 pass. The only 3 failures are unchanged `PreferencesResourceTest` path-handling cases caused by this checkout's non-ASCII parent directory. Exact upstream `main` reproduces all 3 under the same non-ASCII path, while all 3 pass from an ASCII `/tmp` worktree.
- Container integration tests could not run locally because the Docker daemon is unavailable; the attempted failsafe run stopped only on Testcontainers environment discovery.
- Independent review: GO, 0 Critical / 0 Important; one non-blocking Minor noted that the `both` unit case could additionally pin the combined message ordering.

No equivalent open PR was found, the branch is 0 behind / 1 ahead, and no tracked build artifacts changed.

